PR #14, opened 2020-06-17
Add AWS Cognito authentication component

bchapuis:
This component lets user login with AWS Cognito. I would love to get general feedbacks on this pull request. Could you also provide some hints on the following points:
- LocalStorage is used to store a state variable that needs to survive the redirections performed when the user login. The use of this variable is recommended to prevent csrf. 
- The OAuth2 parameters are set in the index.html file and would I love to now if you already have a mean to store configuration variables. In particular, the redirection url should be set according to the deployment environment.
- I wanted to write some unit tests for the initUser function. Where should I put them?


gberaudo:
@bchapuis,

>    LocalStorage is used to store a state variable that needs to survive the redirections performed when the user login. The use of this variable is recommended to prevent csrf.

Yes you can use localstorage if you need to.

>    The OAuth2 parameters are set in the index.html file and would I love to now if you already have a mean to store configuration variables. In particular, the redirection url should be set according to the deployment environment.

If this can be computed using the hostname and prefix (reviews have an extra level with the PR name) then do it directly inside your component.

>    I wanted to write some unit tests for the initUser function. Where should I put them?

It is a good idea but we have no way to do that yet.

bchapuis:
@gberaudo @fredj Great, thank you for the responses. Using the hostname and prefix to generate the redirection URL is a good idea.

Does it bother you if we have the following variables hard coded in the `index.js` file? Some of these variables may change depending on the deployment environment.

```js
// setup the OAuth2 parameters
auth.endpoint = 'https://mylogin.auth.eu-central-1.amazoncognito.com/oauth2/authorize';
auth.responseType = 'token';
auth.clientId = '5k1mgef7ggiremt415eecn95ki';
auth.redirectUri = 'http://localhost:8000/';
auth.scope = 'openid+profile';
```

fredj:
> @gberaudo @fredj Great, thank you for the responses. Using the hostname and prefix to generate the redirection URL is a good idea.
> 
> Does it bother you if we have the following variables hard coded in the `index.js` file? Some of these variables may change depending on the deployment environment.
> 
> ```js
> // setup the OAuth2 parameters
> auth.endpoint = 'https://mylogin.auth.eu-central-1.amazoncognito.com/oauth2/authorize';
> auth.responseType = 'token';
> auth.clientId = '5k1mgef7ggiremt415eecn95ki';
> auth.redirectUri = 'http://localhost:8000/';
> auth.scope = 'openid+profile';
> ```

I've changed it in 74a0e7d7b3f371181076435e891a3c9940012d04: the default values are now in `ngm-auth`

fredj:
As I said privately to @bchapuis; we could change the code to mimic the https://github.com/pwa-builder/pwa-auth web component.
Instead of calling `initUser` in `index.js`; `ngm-auth` could do it and fire an event when it's done. Something like:
```js
const auth = document.querySelector('ngm-auth');
auth.addEventListener('signin-completed', event => {
  const signIn = event.detail;
  if (signIn.user) {
    // ...
  }
});
```

fredj:
> @fredj the endpoint property is similar to the clientId property. The mylogin and region part of the url depends on our configuration of cognito. They both may change depending on our deployment environment (test, dev, prod).

Done, see 7772b7e

bchapuis:
Following the discussion with @fredj, we will investigate the use of a popup to authenticate instead of redirecting the current page. This could also help in addressing the permalink problem (the redirection url which contains the hash gets rewritten).

bchapuis:
@gberaudo regarding the tests, do you have a preference (mochajs, jest, etc.)?

gberaudo:
@bchapuis, for the tests, the easiest one.

gberaudo:
maybe you can check if `node -r esm` is helping with KISS.

gberaudo:
@bchapuis, cool the CI is now triggering after you merged in the master branch.
Instead of mergind, can you rebase on top of master?

bchapuis:
@gberaudo and the authentication can now be [tested](https://review.swissgeol.ch/oauth2_GSNGM-249/).

I did a rebase initially and it also resulted merge conflicts. As some of my changes were a bit exploratory, an idea could be to squash them at the end. What do you think?

webpack.config.js
```diff
@@ -34,40 +34,42 @@ module.exports = smp.wrap({
...
       },
     ],
   },
-
   devServer: {
     contentBase: path.join(__dirname, 'dist'),
     compress: true,
-    port: 8000
+    port: 8000,
+    watchOptions: {
```
gberaudo, line 49: Why do you need to change to polling mode? It is supposed to work out of the box: https://webpack.js.org/configuration/watch/#watchoptionspoll

src/elements/ngm-auth.js
```diff
@@ -71,22 +19,41 @@ class NgmAuth extends LitElement {
...
 
   logout() {
-    this.user = undefined;
+    Auth.removeUser();
+    this.user = Auth.getUser();
   }
 
   render() {
-    if (this.user === undefined) {
+    if (this.user === null) {
```
gberaudo, line 60:
```suggestion
    if (!this.user) {
```

src/auth.js
```diff
@@ -0,0 +1,79 @@
+
+const cognitoState = 'cognito_state';
+const cognitoUser = 'cognito_user';
+
+const isResponse = /^#[\w]+=[\w.=-]+(&[\w]+=[\w.=-]+)*$/;
+const isToken = /^[\w=-]+.[\w=-]+.[\w=-]+$/;
```
gberaudo, line 6: Could you add in comment 2 examples of response / token?

src/auth.js
```diff
@@ -0,0 +1,79 @@
...
+    static setUser(user) {
+        const value = JSON.stringify(user);
+        localStorage.setItem(cognitoUser, value);
+    }
+
+    static removeUser() {
+        localStorage.removeItem(cognitoUser);
+    }
+
+    static async authenticate() {
```
gberaudo, line 71: There is no code doing the authentication. Shouldn't it be named `waitForAuthentication` instead?

test/auth.test.js
```diff
@@ -0,0 +1,87 @@
...
+// initialize some constants
+const url = 'http://localhost/';
+const user = {name: "John Doe"};
+const payload = Buffer.from(JSON.stringify(user)).toString('base64').replace(/=/g, '');
+const jwt = `header.${payload}.signature`;
+const token = `#access_token=${jwt}`;
+const type = '&token=Bearer';
+const state = `&state=test`;
+
+// initilize the window, document and localStorage objects
```
gberaudo, line 13:
```suggestion
// initialize the window, document and localStorage objects
```

test/auth.test.js
```diff
@@ -0,0 +1,87 @@
...
+      assert.deepEqual(Auth.getUser(), user);
+      Auth.removeUser();
+      assert.deepEqual(Auth.getUser(), null);
+    });
+  });
+
+  describe('authenticate', function () {
+    it('should wait until the user authenticates', async function () {
+      Auth.removeUser()
+      setInterval(() => Auth.setUser(user), 100);
```
gberaudo, line 46: Let's use something a bit bigger to test the `while` loop.
```suggestion
      setInterval(() => Auth.setUser(user), 120);
```

test/auth.test.js
```diff
@@ -0,0 +1,87 @@
...
+      assert.throws(() => Auth.parseToken(undefined), Error);
+      assert.throws(() => Auth.parseToken(undefined), Error);
+    });
+    it('should parse a well formed input', function () {
+      assert.deepEqual(Auth.parseToken(jwt), user);
+    });
+  });
+
+  describe('initialize', function () {
+    it('should throw an error when the input is input', function () {
```
gberaudo, line 79: What should it do?

gberaudo:
@bchapuis, yes squashing all is OK.

gberaudo:
(and great to have been able to add tests!)

webpack.config.js
bchapuis: The automatic reload of the app doesn't work on my desktop computer. This fixed the issue.

test/auth.test.js
bchapuis: Thanks ;)

bchapuis:
We could easily use a popup instead of a tab but I'm not sure about the final settings for eAIM (width, height, etc.)

package.json
```diff
@@ -7,7 +7,9 @@
     "start": "npm run webpack-start",
     "webpack-start": "node scripts/environment.js && node scripts/generateManuals.js && webpack-dev-server",
     "rollup-start": "node scripts/environment.js && node scripts/generateManuals.js && rollup --environment SERVE:true -c --watch --no-treeshake",
-    "build": "node scripts/environment.js $RELEASE_NAME && node scripts/generateManuals.js && rm -rf dist/* && rollup --environment mode:production -c && scripts/create_versions.sh"
+    "build": "node scripts/environment.js $RELEASE_NAME && node scripts/generateManuals.js && rm -rf dist/* && rollup --environment mode:production -c && scripts/create_versions.sh",
+    "test": "mocha",
+    "test-watch": "mocha --watch"
```
fredj, line 12: minor but instead of adding `test-watch` it can be done by adding `--` to `test`:
```
npm run test -- --watch
```
I do the same for the `lint` script:
```
npm run lint -- --fix
```

danduk82:
@fredj  is it OK to merge for you?

PR #50, opened 2020-06-24
Full screen view

vladyslav-tk:


fredj:
@vladyslav-tk is there any reason why the [Fullscreen_API](https://developer.mozilla.org/en-US/docs/Web/API/Fullscreen_API) is not used?

vladyslav-tk:
> @vladyslav-tk is there any reason why the [Fullscreen_API](https://developer.mozilla.org/en-US/docs/Web/API/Fullscreen_API) is not used?

this API works the same as 'F11' button but as I understood from issue description, full-screen view means without interface (left sidebar, navigation buttons and etc.). I will clarify this with Elisabeth

vladyslav-tk:
> > @vladyslav-tk is there any reason why the [Fullscreen_API](https://developer.mozilla.org/en-US/docs/Web/API/Fullscreen_API) is not used?
> 
> this API works the same as 'F11' button but as I understood from issue description, full-screen view means without interface (left sidebar, navigation buttons and etc.). I will clarify this with Elisabeth

it will be combination of what implemented now and fullscreen API

PR #63, opened 2020-07-01
Allow to set attribute order for each layer

vladyslav-tk:


src/objectInformation.js
```diff
@@ -2,12 +2,17 @@ import JulianDate from 'cesium/Source/Core/JulianDate';
...
+  let propertyNames = primitive.getPropertyNames();
   const length = propertyNames.length;
+  const properties = primitive.tileset.properties;
+  const propsOrder = properties && properties.propsOrder ? properties.propsOrder : [];
+  propertyNames = sortPropertyNames(propertyNames, propsOrder);
   for (let i = 0; i < length; ++i) {
     const key = propertyNames[i];
     const value = primitive.getProperty(key);
-    data[key] = value;
+    if (value) {
```
fredj, line 13: warning: it will hide the property if the value is `0`

vladyslav-tk: fixed

PR #68, opened 2020-07-02
CH-Logo as Home Button

vladyslav-tk:


src/index.js
```diff
@@ -165,3 +165,7 @@ i18next.on('initialized', () => {
   showSlowLoadingWindow();
 });
 
+const origin = window.location.origin;
+const pathname = window.location.pathname;
+document.querySelector('#ngm-home-link').href = origin.includes('review') ? `${origin}${pathname}` : origin;
```
fredj, line 170: can it be simplified to:
```suggestion
document.querySelector('#ngm-home-link').href = `${origin}${pathname}`;
```
If it's not `review` `pathname` is `/` and the url is something like `https://dev.swissgeol.ch/` (which is correct)

vladyslav-tk: updated

PR #69, opened 2020-07-02
Visualisation | Bounding box - depending on layer config

vladyslav-tk:


fredj:
I'm wondering if the button should be set to `disabled` instead of being hidden. We could change it later if needed

PR #84, opened 2020-07-15
Geocover query

gberaudo:


src/query/QueryManager.js
```diff
@@ -0,0 +1,63 @@
...
+      if (popupContent) {
+        popupContent = popupContent.replace(/cell-left/g, 'key')
+        .replace(/<td>/g, '<td class="value">')
+        .replace(/<table>/g, '<table class="ui compact small very basic table">');
+      }
+      const onshow = () => {
+        console.log('showing', d.geometry);
+      };
+      const onhide = () => {
+        console.log('hiding', d.geometry);
```
vladyslav-tk, line 35: Is console.log required?

src/query/QueryManager.js
```diff
@@ -0,0 +1,63 @@
...
+      };
+      return {
+        popupContent,
+        onhide,
+        onshow
+      };
+    }
+  }
+
+  async onclick(click) {
```
vladyslav-tk, line 45: There are some issues with a click:

1. When clicking out of map it returns the error: https://i.imgur.com/wtJvdN6.jpeg
2. When drawing popup appears: https://i.imgur.com/dY8oiTW.jpeg

gberaudo: I fixed these, thanks.

vladyslav-tk: You disabled query when tools are open and now AOI cant be selected when AOI section is open

PR #89, opened 2020-07-17
Query on geo.admin geocover layer - Highlight on the 3d viewer

vladyslav-tk:


src/query/QueryManager.js
```diff
@@ -59,4 +68,50 @@ export default class QueryManager {
...
+        polyline: {
+          positions: convertedCoords,
+          material: OBJECT_HIGHLIGHT_COLOR
+        }
+      });
+    }
+    this.scene.requestRender();
+  }
+
+  unhiglight() {
```
gberaudo, line 110: unhighlight

src/query/QueryManager.js
```diff
@@ -2,15 +2,23 @@ import ObjectSelector from './ObjectSelector';
...
   constructor(viewer) {
     this.objectSelector = new ObjectSelector(viewer);
     this.swisstopoIndentify = new SwisstopoIdentify();
+    this.viewer = viewer;
     this.scene = viewer.scene;
     this.enabled = true;
+    this.highlightEntity = null;
     viewer.screenSpaceEventHandler.setInputAction(click => this.onclick(click), ScreenSpaceEventType.LEFT_CLICK);
+    document.querySelector('ngm-object-information')
+      .addEventListener('closed', () => this.unhiglight());
```
gberaudo, line 21: I think we should not do global DOM queries in components. It breaks the encapsulation of the component which is going against the concept of components.

I suggest instead to create a "onclose" function attribute that must be called by the 'ngm-object-information" component when it closes. If you remember, it is what I had started doing in my previous PR and removed for the merging.

src/query/QueryManager.js
```diff
@@ -59,4 +68,50 @@ export default class QueryManager {
...
+
+        const ent = new Entity({
+          polygon: {
+            hierarchy: convertedCoords,
+            material: OBJECT_HIGHLIGHT_COLOR.withAlpha(0.7)
+          }
+        });
+        entity.merge(ent);
+      });
+      this.viewer.entities.add(entity);
```
gberaudo, line 92: Great!

src/query/QueryManager.js
```diff
@@ -59,4 +68,50 @@ export default class QueryManager {
...
+      });
+      this.highlightEntity = this.viewer.entities.add({
+        polyline: {
+          positions: convertedCoords,
+          material: OBJECT_HIGHLIGHT_COLOR
+        }
+      });
+    }
+    this.scene.requestRender();
+  }
```
gberaudo, line 108: For other geometries please do a console.error so that we can easily detect and fix the issue.

PR #90, opened 2020-07-20
add explicit config file for aws-cli

danduk82:


scripts/deploy_to_s3.sh
```diff
@@ -43,6 +43,22 @@ then
     exit 1
 fi
 
+# if no AWS config exists, create it.
+if [[ ! -d ~/.aws ]] ; then
+  mkdir ~/.aws
+fi
+
+if [[  ! -r ~/.aws/config ]] || [ -z "$(grep ngmdeploy ~/.aws/config)" ] ; then
```
gberaudo, line 51: To be extra safe about messing up with the local dev environment I would check for the presence of a
MESS_WITH_AWS_CONFIG=true environment variable (that you set in the workflow of the actions).

Otherwise LGTM.

PR #120, opened 2020-08-12
Add script to fetch all the 3dtiles properties

fredj:


locales/de.json
```diff
@@ -1,47 +1,109 @@
 {
   "aare_valley_label": "Aaretal",
+  "access": "",
```
gberaudo, line 3: I feel bad having both application translations and data translations in the same file.
Have you considered using i18next namespaces? It looks like designed exactly for our use-case: https://www.i18next.com/principles/namespaces

Concretly the data translations would be put in their own json and will be loaded by an extra fetch.

scripts/get_assets_properties.js
```diff
@@ -0,0 +1,40 @@
...
+  });
+}
+
+const header = `
+/* eslint-disable */
+// DO NOT EDIT: this is file is automatically generated by get_assets_properties script
+`;
+
+// get all the user's assets
+const endpoints = fetchIon('https://api.cesium.com/v1/assets')
```
gberaudo, line 19: Cesium ION contains assets not used in production.
A better way is to import the `defaultLayerTree` from the `constants.js` file:
-  it is always up-to-date;
- only reference used Cesium ION assets;
- also reference non-Cesium ION assets.

locales/de.json
fredj: That's a very good idea

scripts/get_assets_properties.js
fredj: I'm trying with this `.mjs` file:
```js
import {defaultLayerTree} from '../src/constants.js';
```
but the error is:
```
(node:786836) ExperimentalWarning: The ESM module loader is experimental.
file:///home/fredj/code/ngm/scripts/boo.mjs:1
import {defaultLayerTree} from '../src/constants.js';
        ^^^^^^^^^^^^^^^^
SyntaxError: The requested module '../src/constants.js' is expected to be of type CommonJS, which does not support named exports. CommonJS modules can be imported by importing the default export.
For example:
import pkg from '../src/constants.js';
const {defaultLayerTree} = pkg;
```
So I've tried with:
```js
import constants from '../src/constants.js';
const {defaultLayerTree} = constants;
```
error is:
```
(node:787164) ExperimentalWarning: The ESM module loader is experimental.
/home/fredj/code/ngm/src/constants.js:1
import Rectangle from 'cesium/Source/Core/Rectangle';
^^^^^^

SyntaxError: Cannot use import statement outside a module
```

package.json
```diff
@@ -3,6 +3,7 @@
   "private": true,
   "scripts": {
     "lint": "eslint src/ test/",
+    "extract-from-assets": "node scripts/get_assets_properties.js > src/assets_properties.js",
```
gberaudo, line 6: To fix the error below, have you tried with `node -r esm scripts/get_assets_properties.js > src/assets_properties.js`?

fredj: the error is now:
```
/home/fredj/code/ngm/node_modules/cesium/Source/Core/Rectangle.js:1
Error [ERR_REQUIRE_ESM]: Must use import to load ES Module: /home/fredj/code/ngm/node_modules/cesium/Source/Core/Rectangle.js
require() of ES modules is not supported.
require() of /home/fredj/code/ngm/node_modules/cesium/Source/Core/Rectangle.js from /home/fredj/code/ngm/src/constants.js is an ES module file as it is a .js file whose nearest parent package.json contains "type": "module" which defines all .js files in that package scope as ES modules.
Instead rename Rectangle.js to end in .cjs, change the requiring code to use import(), or remove "type": "module" from /home/fredj/code/ngm/node_modules/cesium/package.json.

    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1153:13) {
  code: 'ERR_REQUIRE_ESM'
}
```

gberaudo: @fredj, I pushed a workaround to fix this. You can continue from there.

fredj: Thanks ! I started with the same idea yesterday but didn't finished

fredj: It works but the `export` keyword is not supported by node:
```
export const LAS_POINT_CLOUD_STYLE = {
^^^^^^

SyntaxError: Unexpected token 'export'
    at wrapSafe (internal/modules/cjs/loader.js:1053:16)
    at Module._compile (internal/modules/cjs/loader.js:1101:27)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1157:10)
    at Module.load (internal/modules/cjs/loader.js:985:32)
    at Function.Module._load (internal/modules/cjs/loader.js:878:14)
    at ModuleWrap.<anonymous> (internal/modules/esm/translators.js:155:15)
    at ModuleJob.run (internal/modules/esm/module_job.js:139:37)
    at async Loader.import (internal/modules/esm/loader.js:179:24)
```

fredj:
@gberaudo it's finally working !

PR #144, opened 2020-08-25
Avoid NaN height in ngm-feature-height

fredj:
`getHeight` can return `undefined` and `NaN` is displayed


gberaudo (approved):
You can also do:
```js
const altitude = this.viewer.scene.globe.getHeight(position) || 0;
```

PR #150, opened 2020-08-28
use non priviledged terrain from swisstopo

danduk82:
This fixes the VIP blocked terrain issue

### DIFF
--- a/src/viewer.js
+++ b/src/viewer.js
@@ -93,7 +93,7 @@ export function setupViewer(container) {
   let terrainUrl;
   const ownTerrain = searchParams.get('ownterrain') !== 'false';
   if (ownTerrain) {
-    terrainUrl = 'https://terrain100.geo.admin.ch/1.0.0/ch.swisstopo.terrain.3d';
+    terrainUrl = 'https://3d.geo.admin.ch/1.0.0/ch.swisstopo.terrain.3d/default/20180601/4326/';
   } else {
     terrainUrl = IonResource.fromAssetId(1);
   }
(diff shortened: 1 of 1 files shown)


src/viewer.js
```diff
@@ -93,7 +93,7 @@ export function setupViewer(container) {
   let terrainUrl;
   const ownTerrain = searchParams.get('ownterrain') !== 'false';
   if (ownTerrain) {
-    terrainUrl = 'https://terrain100.geo.admin.ch/1.0.0/ch.swisstopo.terrain.3d';
+    terrainUrl = 'https://3d.geo.admin.ch/1.0.0/ch.swisstopo.terrain.3d/default/20180601/4326/';
```
gberaudo, line 96: It does not look correct to use this old timestamp.

danduk82: it is the one officially advertised on api.geo.admin.ch

danduk82: we should summon and ask to @rebert

danduk82: I will send him a mail

gberaudo: Here is the URL/timestamp used by geoadmin:

GET | https://terrain101.geo.admin.ch/1.0.0/ch.swisstopo.terrain.3d/default/20200520/4326/8/269/194.terrain?v=1.0.0

It is also the one Tobias mentionned in the Cesium ticket back then.

danduk82: @gberaudo  in another NGM ticket I found a new timestamp: 20200520/4326/
I tried it and it seems to work

What do you think?

danduk82: seems the same to me

fredj (approved):

PR #166, opened 2020-09-01
slice according to camera center

vladyslav-tk:


vladyslav-tk:
@fredj this one possibly not actual anymore. I added planes with a possibility to move in https://github.com/swissgeol/ngm/pull/168

PR #236, opened 2020-10-27
GSNGM-400

vladyslav-tk:


fredj:
The corners are not precisely aligned:
screenshot_20201029_091004

Can you try to fix it?

vladyslav-tk:
@fredj I made an investigation but I have no solution to this. I think it happens because a plane is always 
straight line but the map is not rectangular and on long planes this difference becomes visible

fredj (approved):

fredj:
> @fredj I made an investigation but I have no solution to this. I think it happens because a plane is always
straight line but the map is not rectangular and on long planes this difference becomes visible

ok, please merge

PR #247, opened 2020-10-30
[not working] Use of the token endpoint

gberaudo:
Unfortunately it does not work.
I always get a 400 error saying the client id is incorrect.

Infos:
- the app now uses a different cognito user pool with only one user: me :P
- this should be changed back in order to test with another user;

Useful resources:
- to inspect JWT tokens: https://jwt.io/
- token enpoint doc: https://docs.aws.amazon.com/cognito/latest/developerguide/token-endpoint.html

### DIFF
--- a/src/auth.js
+++ b/src/auth.js
@@ -17,6 +17,28 @@ export default class Auth {
   static initialize() {
     // try parse and store the cognito response
     // and fail silently otherwise
+    const params = new URLSearchParams(window.location.search);
+    if (params.has('code')) {
+      const code = params.get('code');
+      const state = params.get('state');
+      localStorage.setItem('mycode', window.location.search);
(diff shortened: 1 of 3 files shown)


src/auth.js
```diff
@@ -17,6 +17,28 @@ export default class Auth {
...
+      const code = params.get('code');
+      const state = params.get('state');
+      localStorage.setItem('mycode', window.location.search);
+      const endpoint = 'https://mysecurelogin.auth.eu-central-1.amazoncognito.com/oauth2/token';
+      const clientId = '16osqbbqrstpo8tjf94st0nrpg';
+      const data = new URLSearchParams();
+      data.append('grant_type', 'authorization_code');
+      data.append('code', code);
+      //data.append('client_id', clientId);
+      data.append('redirect_url', 'http://localhost:8000/');
```
danduk82, line 31: @gberaudo  I would put something like `      data.append('redirect_url', location.origin);` instead

PR #265, opened 2020-11-04
Add ngm-loading-mask element

fredj:


src/elements/ngm-loading-mask.js
```diff
@@ -0,0 +1,52 @@
...
+  updated() {
+    if (this.step > 0) {
+      this.progress.progress('set percent', this.step / this.total * 100);
+    }
+  }
+
+  render() {
+    return html`
+      <div class="ui ${this.active ? 'active' : ''} inverted dimmer">
+        <div class="ui blue tiny progress ${this.step === 0 ? 'sliding indeterminate' : ''}">
```
gberaudo, line 40: There is https://lit-html.polymer-project.org/guide/template-reference#classmap to handle this.
But I don't know the advantage compared to your simple condition.

gberaudo (approved):

PR #266, opened 2020-11-05
Logout on token expiration

gberaudo:


src/auth.js
```diff
@@ -4,32 +4,29 @@ const cognitoState = 'cognito_state';
...
+        this.setAccessToken(params.get('id_token'));
       }
-    } catch (e) {
-      // do nothing
     }
 
     const accessToken = this.getAccessToken();
     if (accessToken) {
-      _AWSCredentials = fromCognitoIdentityPool({
+      window['AWSCred'] = _AWSCredentials = fromCognitoIdentityPool({
```
fredj, line 29:
```suggestion
      _AWSCredentials = fromCognitoIdentityPool({
```
I guess it was only for the debug

src/elements/ngm-auth.js
```diff
@@ -25,12 +25,32 @@ class NgmAuth extends I18nMixin(LitElement) {
...
     this.redirectUri = `${location.origin}${location.pathname}`;
     this.scope = 'openid+profile';
   }
 
+  /**
+   * Set or clear auto logout timer.
+   * @param claims
+   */
+  updateLogoutTimeout_(claims) {
+    if (this.expireTimer_) {
```
fredj, line 39: will it work ? shouldn't `this.expireTimer_` be set to `null` after `clearTimeout(this.expireTimer_);` ?

src/auth.js
```diff
@@ -4,32 +4,29 @@ const cognitoState = 'cognito_state';
...
-      if (response.token_type === 'Bearer' && response.state === this.state()) {
-        this.setUser(this.parseToken(response.access_token));
-        this.setAccessToken(response.id_token);
+    if (window.location.hash.startsWith('#')) {
+      // https://docs.aws.amazon.com/cognito/latest/developerguide/amazon-cognito-user-pools-using-tokens-with-identity-providers.html
+      const response = window.location.hash.substring(1);
+      const params = new URLSearchParams(response);
+      if (params.has('access_token') && params.has('id_token') &&
+          params.get('token_type') === 'Bearer' && params.get('state') === this.state()) {
+        localStorage.setItem('rawCognitoResponse', response);
```
fredj, line 18: minor but `localStorage.removeItem('rawCognitoResponse');` could be added in `logout`

fredj: but the item is never read, remove this line ?

src/auth.js
gberaudo: Andrea said he did it on purpose in its code. So I restored it.

src/auth.js
gberaudo: It is for debugging. The login process is very complicate and envolve opening and closing browser windows.
As a consequence it is not easy to see the cognito response (breakpoints do not work) when debugging, except by storing it in the local storage.

fredj (approved):

PR #279, opened 2020-11-16
GSNGM-407

vladyslav-tk:


src/utils.js
```diff
@@ -244,3 +244,38 @@ export function applyLimits(value, minValue, maxValue) {
...
+
+/**
+ * Returns point of lines intersection or null
+ * @param {Cartesian2} start1
+ * @param {Cartesian2} end1
+ * @param {Cartesian2} start2
+ * @param {Cartesian2} end2
+ * @return {Cartesian2|null}
+ */
+export function getLinesIntersection(start1, end1, start2, end2) {
```
gberaudo, line 256: See https://cesium.com/docs/cesiumjs-ref-doc/Intersections2D.html?classFilter=interse#.computeLineSegmentLineSegmentIntersection

vladyslav-tk: replaced custom function with cesium one

gberaudo:
@vladyslav-tk, I do not see the part about the rotation handle. It is showing on your side?:
image

vladyslav-tk:
@gberaudo yes, It shows: https://i.imgur.com/XKqXJm7.png 
but this is strange that you can't see it

vladyslav-tk:
@gberaudo fixed. It was reproducible only in firefox because firefox doesn't support SVG without width/height

gberaudo:
Indeed, I see it now @vladyslav-tk. But there is something strange:
- the symbol is visible through the terrain (no detph test);
- it does not really follow the mouse (when you drag it, there is quickly some large distance between the symbol and the cursor).

vladyslav-tk:
@gberaudo 

> the symbol is visible through the terrain (no detph test);
yes, I disabled the depth test because I thought it should be visible always

>it does not really follow the mouse (when you drag it, there is quickly some large distance between the symbol and the cursor)
Do you mean like this?: https://i.imgur.com/NUPEQGD.gif

gberaudo:
> > the symbol is visible through the terrain (no detph test);
> yes, I disabled the depth test because I thought it should be visible always

It makes it hard do understand where we click/drag, but OK since it is also the behaviour of the rectangle corners.

> > it does not really follow the mouse (when you drag it, there is quickly some large distance between the symbol and the cursor)
> Do you mean like this?: https://i.imgur.com/NUPEQGD.gif
Yes

gberaudo (approved):

PR #291, opened 2020-11-23
GSNGM-434: Open link in new tab/window

b4l:


src/elements/ngm-object-information.js
```diff
@@ -46,7 +46,7 @@ class NgmObjectInformation extends I18nMixin(LitElement) {
             return html`
           <tr class="top aligned">
             <td class="key">${i18next.t(key)}</td>
-            <td class="val"><a href="${value}">${value}</a></td>
+            <td class="val"><a href="${value}" target="_blank">${value}</a></td>
```
gberaudo, line 49:
```suggestion
            <td class="val"><a href="${value}" target="_blank" rel="noopener">${value}</a></td>
```

gberaudo (approved):

src/elements/ngm-object-information.js
b4l: thx!

PR #293, opened 2020-11-25
Add tileset progress bar

gberaudo:


src/elements/ngm-left-side-bar.js
```diff
@@ -36,6 +36,7 @@ class LeftSideBar extends I18nMixin(LitElement) {
       hideWelcome: {type: Boolean},
       mapChooser: {type: Function},
       authenticated: {type: Boolean},
+      globeQueueLength_: {type: Number},
```
fredj, line 39: instead of having `globeQueueLength_` here it could be a "standard" variable and `this.requestUpdate()` called when the value is changed

vladyslav-tk:
@gberaudo it looks nice

Some things that I noticed:
1. Need to align a little bit: https://i.imgur.com/T5GAL9c.png
2. Seems like currently, spinner for tiles shows only on page load and not when you add a new layer: https://i.imgur.com/sarITrz.gif

src/elements/ngm-layer-loading-indicator.js
```diff
@@ -0,0 +1,38 @@
+import {LitElement, html} from 'lit-element';
+
+
+class NgmLayerLoadingIndicator extends LitElement {
+
+  static get properties() {
+    return {
+      loading: {type: Number},
+      config: {type: Object},
+      actions: {type: Number}
```
fredj, line 10:
```suggestion
      actions: {type: Object}
```

gberaudo:
>     2. Seems like currently, spinner for tiles shows only on page load and not when you add a new layer: https://i.imgur.com/sarITrz.gif

Indeed, there is something very strange here. I added some debugging and the newly added layer triggers an addEventListener on a wrong tileset (it takes the layer at the bottom of the list). I will not be working this afternoon, if someone wants to dig into this it is welcomed.

gberaudo:
@vladyslav-tk, @fredj, while working on this feature I noticed the side bar (and all the active layers) are immediately downloaded by CesiumJS, even during the step 1. I have reworked the code to fix this and it looks to me the loading is much faster now (with ownterrain=false).

vladyslav-tk (approved):

src/elements/ngm-left-side-bar.js
gberaudo: It is is a difficult decision:
- on an API point of view, we should not expose private properties to the outside;
- on a clarity point of view it makes more sense to let lit-element trigger renders (and also clearly define what are the dependencies of the template).

I would stay with the 2nd option. And maybe we can discuss this with the lit-element people, there are currently thinking about lit-element 3.x.

src/elements/ngm-left-side-bar.js
```diff
@@ -352,13 +363,20 @@ class LeftSideBar extends I18nMixin(LitElement) {
   }
 
   accordionFactory(element) {
+    console.log(element, element.id);
```
vladyslav-tk, line 366: @gberaudo we missed console.log here

PR #296, opened 2020-11-26
GSNGM-385: New borehole data

b4l:


gberaudo (approved):
What about the new translation?

b4l:
> What about the new translation?

It is specified to have this new attribute on the stick of the boreholes, so i added it.

PR #311, opened 2020-12-03
Allow using cli terrain

gberaudo:


gberaudo:
Merging so that it can be tested by Swisstopo partners.

PR #363, opened 2021-01-11
Translations fix for popup and dropdown (GSNGM-468)

vladyslav-tk:


src/toolbox/ngm-gst-interaction.js
```diff
@@ -65,11 +64,12 @@ class NgmGstInteraction extends LitElementI18n {
   }
 
   firstUpdated() {
-    $(this.querySelector(`#${createBtnId}`)).popup({
+    $(this.querySelector('.ngm-create-section')).popup({
       position: 'top left',
       content: i18next.t('tbx_create_section_hint'),
       variation: 'mini',
       onShow: () => {
+        $(this.querySelector('.ngm-create-section')).popup('change content', i18next.t('tbx_create_section_hint'));
```
gberaudo, line 72: Do you have a link to some documentation about this 'change content' functionnality? I could not find anything in the fomantc-ui official doc.

vladyslav-tk: See 'Behavior' - https://fomantic-ui.com/modules/popup.html#behavior

gberaudo: Thanks.

Another approach maybe more straightforward would be to use a lit-element template as the initial "content".
That way the popup content would be automatically changed as soon as the user swich language and the code would stay easy to understand.

We could also create an ngm-popup element that would expose a translatable fomantic-ui popup.

fredj: maybe something like: https://github.com/swissgeol/ngm/commit/09067a9ec26a121b21b5a1e5550612883159a052

vladyslav-tk: yes, looks good. It will be useful for a popup on icons. For the button, we can use `data-tooltip` attribute (in the case of gst, jQuery added just to show extent but it can be added separately) but for icons, this attribute doesn't work. I will merge your commit in this PR

fredj (approved):

PR #364, opened 2021-01-12
convert to read-only

b4l:


gberaudo (approved):

b4l:
Instead of changing we will restrict the current token.

PR #369, opened 2021-01-14
Downlaod ticked geometries as one single KML (GSNGM-463)

vladyslav-tk:


src/utils.js
```diff
@@ -122,3 +122,17 @@ export function filterCsvString(str, bbox4326) {
   const filteredString = lines.join('\n');
   return filteredString;
 }
+
+/**
+ * Check is string can be parsed to object
+ * @param string
```
fredj, line 128:
```suggestion
 * @param {string} string
```

src/cesiumutils.js
```diff
@@ -218,3 +218,26 @@ export function cartesianToDegrees(cartesian) {
...
 }
+
+/**
+ * Extend kml for export with entities properties
+ * @param {string} kml - kml for export
+ * @param {Array<Entity>} entities - list of entities for export
+ * @return {string}
+ */
+export function extendKmlWithProperties(kml, entities) {
+  entities.values.forEach(ent => {
```
gberaudo, line 229:
```suggestion
  entities.values.forEach(entity => {
```

src/cesiumutils.js
```diff
@@ -218,3 +218,26 @@ export function cartesianToDegrees(cartesian) {
...
+    let kmlProperties = '<ExtendedData>';
+    ent.properties.propertyNames.forEach(prop => {
+      let value = ent.properties[prop] ? ent.properties[prop].getValue() : undefined;
+      if (value || typeof value === 'number') {
+        value = typeof value === 'object' ? JSON.stringify(value) : value;
```
gberaudo, line 234: I think this is a bit convoluted.
What about something like?
```js
 const valueType = typeof value;
if (valueType === 'number') {
  kmlProperties += `<Data name="${prop}"><value>${value}</value></Data>`;
} else if (valueType === 'object') {
  kmlProperties += `<Data name="${prop}"><value>${JSON.stringify(value)}</value></Data>`;
}
```

And what about string values or arrays?

src/utils.js
```diff
@@ -122,3 +122,17 @@ export function filterCsvString(str, bbox4326) {
...
+ * @returns {boolean}
+ */
+export function isStringJson(string) {
```
gberaudo, line 131: -> `isSerializedJson`?

src/cesiumutils.js
vladyslav-tk: we don't need to check for 'number', it was only for '0' value. Arrays have type 'object'.

I update the condition to make it more clear

src/utils.js
vladyslav-tk: I updated this function. Now, it returns parsed json or undefined to skip the second parse

gberaudo (approved):

PR #393, opened 2021-01-29
Shortlink for swissgeol

vladyslav-tk:


src/elements/ngm-share-link.js
```diff
@@ -0,0 +1,110 @@
...
+import $ from '../jquery';
+import 'fomantic-ui-css/components/popup.js';
+import './ngm-i18n-content.js';
+import {SHORTLINK_HOST_BY_PAGE_HOST} from '../constants';
+
+class NgmShareLink extends LitElementI18n {
+
+  static get properties() {
+    return {
+      shortlink: {type: String}
```
gberaudo, line 14: This is an internal property, so you should be able to use (not tested):
```suggestion
      shortlink: {attribute: false}
```

src/elements/ngm-share-link.js
```diff
@@ -0,0 +1,110 @@
...
+  async getShortlink() {
+    const serviceHost = SHORTLINK_HOST_BY_PAGE_HOST[window.location.host];
+    try {
+      const result = await fetch(`https://${serviceHost}/admin_shrink_url`, {
+        method: 'POST',
+        headers: {
+          'Content-Type': 'application/json;charset=UTF-8'
+        },
+        body: JSON.stringify({
+          'url_long': window.location.href,
```
gberaudo, line 45: It looks like the admin_shrink_url service accepts any URL. Would you be able to accept only our domains.

src/elements/ngm-share-link.js
```diff
@@ -0,0 +1,110 @@
...
+          'cdn_prefix': serviceHost
+        }),
+      });
+      const response = await result.json();
+      if (response && response.url_short) {
+        return response.url_short;
+      }
```
gberaudo, line 52: There is no else. Removing the condition would handle all cases.

src/elements/ngm-share-link.js
```diff
@@ -0,0 +1,110 @@
...
+      return window.location.href;
+    }
+  }
+
+  async onClick() {
+    this.toggleLoader();
+    try {
+      const inputElement = this.querySelector('.ngm-share-link > input');
+      this.shortlink = await this.getShortlink();
+      if (navigator.clipboard) {
```
gberaudo, line 65: Navigator clipboard looks to be supported by all browsers (except IE). What about removing the condition?
https://developer.mozilla.org/en-US/docs/Web/API/Navigator/clipboard#browser_compatibility

src/elements/ngm-share-link.js
```diff
@@ -0,0 +1,106 @@
...
+    }
+  }
+
+  toggleLoader() {
+    const btnElement = this.querySelector('.ngm-share-link > button');
+    const inputElement = this.querySelector('.ngm-share-link > input');
+    btnElement.classList.toggle('disabled');
+    inputElement.disabled = !inputElement.disabled;
+    btnElement.querySelector('.dimmer').classList.toggle('active');
+  }
```
gberaudo, line 84: I think you should not do DOM manipulations like but introduce an internal property `displayLoader`.
Then in the render function you will define the class with the https://lit-html.polymer-project.org/guide/template-reference#classmap directive.

src/elements/ngm-share-link.js
```diff
@@ -0,0 +1,106 @@
...
+
+  render() {
+    return html`
+      <p>${i18next.t('welcome_share_link_label')}</p>
+      <div class="ui mini action input ngm-share-link">
+        <input type="text" readonly value="${this.shortlink}" @click=${this.onClick}>
```
gberaudo, line 90: It is disturbing that the input is updated when ones click it, because you need to click for copying. But I don't have a good idea (apart from removing the input completly :P)

requirements.txt
```diff
@@ -0,0 +1,2 @@
+botocore==1.19.57
+boto3==1.16.47
```
gberaudo, line 2: Why do we now need boto?

src/elements/ngm-share-link.js
vladyslav-tk: Do you mean an update of the lambda function?

src/elements/ngm-share-link.js
vladyslav-tk: We can ask clients about removing it

requirements.txt
vladyslav-tk: I forgot to remove requirements

src/elements/ngm-share-link.js
gberaudo: Yes

gberaudo (approved):

src/elements/ngm-share-link.js
vladyslav-tk: I think we should create an issue for this because it will take some time. Simon propose to make some security key for requests but maybe make host restriction will be more easily

PR #402, opened 2021-02-02
GSNGM-464: Update private and public boreholes

b4l:
Some remarks:
- could not test the private boreholes
- (fixed) `selectTile(feature)` in `QueryManager.js` uses obsolete hard-coded borehole attributes
- attribute translations only exist for the public asset


fredj:
I'm looking into the `selectTile` issue

fredj (approved):

PR #404, opened 2021-02-05
Fix long running text on button

fredj:


src/style/ngm-toolbox.css
```diff
@@ -30,20 +30,16 @@
 }
 
 .ngm-aoi-content .ui.buttons {
-  width: 95%;
+  flex: auto;
```
vladyslav-tk, line 33: Buttons can have a different size because of `flex: auto` - 
image

vladyslav-tk (approved):

PR #406, opened 2021-02-05
Update Cognito endpoint and clientId

ckaenzig:


src/layertree.js
```diff
@@ -133,8 +133,8 @@ const geo_base = {
           visible: false,
           displayed: false,
           restricted: true,
-          aws_s3_bucket: 'ngm-dev-authenticated-resources',
-          aws_s3_key: 'tiles/bh_private_20210201_00/tileset.json',
+          aws_s3_bucket: 'ngm-protected-prod',
+          aws_s3_key: 'tiles/bh_private_20201117_00/tileset.json',
```
fredj, line 137: we should not change this but synchronize the two buckets

PR #446, opened 2021-02-19
Show dimmer on dragenter

fredj:


src/elements/ngm-drop-files.js
```diff
@@ -0,0 +1,66 @@
...
+    return {
+      target: {type: String},
+    };
+  }
+
+  static get styles() {
+    return css`
+
+    `;
+  }
```
gberaudo, line 20: This does not work when there are no shadow DOM.

src/elements/ngm-drop-files.js
```diff
@@ -0,0 +1,66 @@
...
+
+  firstUpdated() {
+    this.dimmer = this.querySelector('.ui.dimmer');
+    const target = document.querySelector(this.target);
```
gberaudo, line 24: Minor: what about passing a DOM element to avoid using the query selector attached to the `document`?

src/elements/ngm-drop-files.js
```diff
@@ -0,0 +1,66 @@
...
+  }
+
+  static get styles() {
+    return css`
+
+    `;
+  }
+
+  firstUpdated() {
+    this.dimmer = this.querySelector('.ui.dimmer');
```
gberaudo, line 23: Can register listeners in the `connectedCallback` and unregister them on `disconnectedCallback`?

src/elements/ngm-drop-files.js
fredj: fixed

src/elements/ngm-drop-files.js
fredj: fixed

src/elements/ngm-drop-files.js
fredj: fixed

src/elements/ngm-drop-files.js
```diff
@@ -0,0 +1,66 @@
...
+  }
+
+  onDragEnter(event) {
+    cancel(event);
+    this.dimmer.classList.add('active');
+  }
+  onDragLeave(event) {
+    if (event.target === this.dimmer) {
+      cancel(event);
+      this.dimmer.classList.remove('active');
```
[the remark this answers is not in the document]
fredj: fixed

PR #457, opened 2021-02-25
Improve csv filtering function

fredj:


src/utils.js
```diff
@@ -105,22 +108,27 @@ export function containsXY(extent, x, y) {
...
+
+  // filter lines
+  const lines = csv.data.filter(line => {
+    const x4326 = +line[x].trim();
+    const y4326 = +line[y].trim();
+    return containsXY(bbox4326, x4326, y4326);
+  });
+
+  // create filtered csv
+  return [
```
gberaudo, line 128: To handle multilines correctly it would be safer to use unparse: https://www.papaparse.com/docs#json-to-csv

fredj: yes, thanks for the hint !

PR #485, opened 2021-03-08
GSNGM-514 switch swissforages URL to prod

eleu:
... I forgot that part: we can already use the prod URL of swissforages!


eleu:
demo works fine!

PR #499, opened 2021-03-12
Slicing-Box size restriction

vladyslav-tk:


src/slicer/SlicingBox.js
```diff
@@ -132,25 +132,33 @@ export default class SlicingBox extends SlicingToolBase {
   }
 
   onPlaneMove(side, moveAmount, moveVector) {
+    let bothSideMove = false;
+    const moveCorners = (position1, position2, oppositePosition1, oppositePosition2) => {
```
fredj, line 136: this function could be move outside of the function

fredj (approved):

PR #535, opened 2021-04-13
Slicing prevents the infobox to open

vladyslav-tk:


src/query/ObjectSelector.js
```diff
@@ -33,15 +38,19 @@ export default class ObjectSelector {
     this.unhighlight();
     let attributes = {};
     if (!object) {
+      const slicerDataSource = this.viewer.dataSources.getByName('slicer')[0];
       const objects = this.scene.drillPick(clickPosition, DRILL_PICK_LIMIT, DRILL_PICK_LENGTH, DRILL_PICK_LENGTH);
       object = objects[0];
+      if (object && object.id && slicerDataSource.entities.contains(object.id)) {
+        object =
```
gberaudo, line 45: Could you add a comment?
Could you also simplify next line by introduction an intermediary variable or using a plain "if" construct?

src/query/ObjectSelector.js
```diff
@@ -78,7 +91,7 @@ export default class ObjectSelector {
     if (aoiDataSource.entities.contains(entity)) {
       const aoiElement = document.querySelector('ngm-aoi-drawer');
       attributes = aoiElement.getInfoProps({...props, name: entity.name});
-    } else if (earthquakesDataSource.entities.contains(entity)) {
+    } else if (earthquakesDataSource && earthquakesDataSource.entities.contains(entity)) {
```
gberaudo, line 94: I think you can use the ".?" operator for that.

gberaudo (approved):

src/query/ObjectSelector.js
vladyslav-tk: We should add an additional loader for this syntaxis

PR #552, opened 2021-04-26
Download cross section data for GeoMol and GeoQuat Aaretal

vladyslav-tk:


b4l:
Properties order is not working for me locally, is this normal?

eleu (changes requested):
if possible, we should make a link of the attribute CS-AAT-Legend, cf. my comment in Jira.

eleu (approved):

PR #564, opened 2021-05-26
Add new WMTS layers from swisstopo

fredj:


src/elements/ngm-left-side-bar.js
```diff
@@ -67,8 +67,7 @@ class LeftSideBar extends LitElementI18n {
     }
 
     this.queryManager.activeLayers = this.activeLayers
-      .filter(config => config.visible)
-      .map(config => config.layer);
+      .filter(config => config.visible && config.queryable !== false);
```
gberaudo, line 70: It is more natural to check for truthy or falsy values. What about introducing a `noQuery: true` attribute?

gberaudo (approved):

src/elements/ngm-left-side-bar.js
fredj: yes, added a new commit

PR #576, opened 2021-06-22
GSNGM-516: Mouse pointer visualization - Integrate the JS code to send information to hotjar

b4l:


gberaudo (commented):
@b4l, shouldn't it be enabled only after the user has given its consent?

b4l:
> @b4l, shouldn't it be enabled only after the user has given its consent?

That seems fair!

gberaudo (approved):

PR #578, opened 2021-06-29
Show legend and geocat link in a popup

fredj:


fredj:
The remove button in the displayed layers doesn't work anymore (working locally)

src/elements/ngm-layer-legend-container.js
```diff
@@ -0,0 +1,24 @@
...
+
+  showLegend(config) {
+    console.assert(config.layer);
+    const legendId = `legend_for_${config.layer}`;
+    if (!this.querySelector('#' + legendId)) {
+      const element = document.createElement('ngm-layer-legend');
+      element.id = legendId;
+      element.config = config;
+      element.addEventListener('close', event => event.target.remove());
+      this.appendChild(element);
```
gberaudo, line 14: It looks strange to do DOM manipulations manually. I would have rather expected the component to maintain a list of "element configs", rendered by lit in the render function.

gberaudo (approved):

src/elements/ngm-layer-legend-container.js
fredj: you're absolutely right; what do you think of 9e9c41b435691926dbe5c9c676d68ed8204ea408 ?

eleu:
I found two issues while testing on the review URL:
- I only get the legend image for "Temperatures - Depth (500mBGL)", for all the others I only have the geocatID. In the Jira issue, the specification contains adding the legend images for all layers in the "Geoenergy" section.
- I get the french legend image although I'm in English / German.

PR #588, opened 2021-07-19
Use solid color for mantel

fredj:


gberaudo (approved):

eleu (approved):
thanks, the color is confirmed by swisstopo as well, so ready to merge also from the PO side!

PR #592, opened 2021-07-28
Set transparent underground color if camera is underground

b4l:


src/viewer.js
```diff
@@ -194,6 +194,14 @@ export function setupViewer(container, rethrowRenderErrors) {
   viewer.scene.postProcessStages.add(fog);
   scene.postRender.addEventListener(scene => {
     fog.enabled = scene.cameraUnderground;
+
+    if (scene.cameraUnderground) {
```
gberaudo, line 198: The condition looks suspicious. Aren't the `undergroundColor*` properties applying only when underground?

b4l: This was from some specific example in the sandcastle. It works perfectly by just setting the globe properties and let cesium do the rest.

vladyslav-tk (approved):

PR #595, opened 2021-08-09
Fix adding layer from search

vladyslav-tk:


gberaudo (approved):
LGTM. In addition, could you rename layer -> layername in this function?

PR #621, opened 2021-09-21
Add basic e2e testing

b4l:


cypress/tsconfig.json
```diff
@@ -0,0 +1,8 @@
+{
+  "compilerOptions": {
+    "target": "es5",
```
gberaudo, line 3: I guess you should not target es5. Maybe es2017 would be good.

cypress/integration/sample_spec.ts
```diff
@@ -0,0 +1,5 @@
+describe('The Home Page', () => {
+  it('sucessfully loads', () => {
+    cy.visit('/');
```
gberaudo, line 3: Could you had one basic check like there is some DOM element we know should be there or something more cool like checking Cesium globe loaded.

b4l: Added declining the tracking consent.

Unfortunately i think what takes long is not DOM but content, also the page has infinite reloads in the cypress test environment and requires a hard coded wait to execute the test more stable.

cypress/tsconfig.json
b4l: Done!

cypress/integration/sample_spec.ts
gberaudo: Hmm, I see. Definitely not something to enable in CI then.
Maybe Cypress is not completly ready for us yet.

gberaudo (approved):

PR #622, opened 2021-09-22
Allow FeatureInfo on Search-added WMTS and WMS

b4l:


src/elements/ngm-left-side-bar.js
```diff
@@ -402,7 +402,8 @@ class LeftSideBar extends LitElementI18n {
       layer: layername,
       visible: true,
       displayed: true,
-      transparency: DEFAULT_LAYER_TRANSPARENCY
+      transparency: DEFAULT_LAYER_TRANSPARENCY,
```
gberaudo, line 405: Looks good for WMTS. Could it also be done for WMS?

b4l: Yes, it should work as is. I can not find where it gets stuck.

b4l: @gberaudo it was actually a precision issue of identify in the case of point and line features. it should be fixed with https://github.com/swissgeol/ngm/pull/622/commits/2470ea0af157cc5a29e680ad5f0b85689fcc88b1

gberaudo (approved):

PR #625, opened 2021-09-24
Set maximumLevel for WMTS layers from swisstopo

fredj:


gberaudo:
@fredj, it does not seem to work.
For geocover, I have errors at zoom level 19 in the console.
But the https://map.geo.admin.ch/configs/en/layersConfig.json shows there should be much more levelz available.
Puzzled...

gberaudo (approved):

PR #626, opened 2021-09-27
Identify all layers

b4l:


src/query/QueryManager.js
```diff
@@ -29,18 +29,17 @@ export default class QueryManager {
 
   async querySwisstopo(pickedPosition, layers) {
     const lang = i18next.language;
-    const identifyData = await this.swisstopoIndentify.identify(pickedPosition, layers, lang);
-    if (identifyData) {
-      const {layerBodId, featureId} = identifyData;
-      let popupContent = await this.swisstopoIndentify.getPopupForFeature(layerBodId, featureId, lang);
+    const identifyResult = await this.swisstopoIndentify.identify(pickedPosition, layers, lang);
+    if (identifyResult) {
+      let popupContent = await this.swisstopoIndentify.getPopupForFeature(identifyResult.layerBodId, identifyResult.featureId, lang);
```
gberaudo, line 34: Why change this?

src/query/QueryManager.js
```diff
@@ -76,10 +75,10 @@ export default class QueryManager {
 
     // we only search the remote Swisstopo service when there was no result for the local search.
     if (attributesEmpty && pickedPosition) {
-      // find the first queryable swisstopo layer
-      const config = this.searchableLayers.find(config => config.queryType === 'geoadmin');
-      if (config) {
-        const result = await this.querySwisstopo(pickedPosition, config.layer);
+      // find all queryable swisstopo layers
+      const layers = this.searchableLayers.filter(config => config.queryType === 'geoadmin').map(config => config.layer);
+      if (layers && layers.length > 0) {
```
gberaudo, line 80: I think `layers` is always defined:
```suggestion
      if (layers.length > 0) {
```

src/query/SwisstopoIdentify.ts
```diff
@@ -1,21 +1,27 @@
...
+  featureId: string;
+}
+
 export default class SwisstopoIdentify {
 
   /**
-   * @param {import ('cesium/Source/Core/Cartesian3.js').default} position
-   * @param {string} layers
+   * @param {Cartesian3} position
+   * @param {Array<string>} layers
```
gberaudo, line 20: We usually write `string[]` but it is also fine like this.

src/query/QueryManager.js
b4l: To align with the interface. I can rename the interface or revert the naming here. Just thought it is more descriptive.

gberaudo: I meant for `const {layerBodId, featureId} = identifyData;`. You did not like ES6 destructuring syntax? ;)

src/query/QueryManager.js
b4l: True, according to the MDN dod!

src/query/SwisstopoIdentify.ts
b4l: Found this way in other files of the project, just copied it. Let's harmonize this another time.

src/query/QueryManager.js
b4l: Oh, I think it is way more expressive without sugar.

gberaudo (approved):

PR #629, opened 2021-10-06
Navigation updates (GSNGM-601, GSNGM-618)

vladyslav-tk:


gberaudo:
@vladyslav-tk, when I drag the red V camera, I sometime see the upload KML UI. Same when using the nav buttons or other widgets. Do you think you can fix it in this PR? Otherwise let's create an issue for next sprint.

PR #652, opened 2021-10-28
GSNGM-621: Camera target point

vladyslav-tk:


eleu:
cool! for the case when the user has drag and dropped the point elsewhere: can we move the camera quickly but smoothly (instead of just jumping) towards the new target point?

vladyslav-tk:
> cool! for the case when the user has drag and dropped the point elsewhere: can we move the camera quickly but smoothly (instead of just jumping) towards the new target point?

there is no direct way to do it. I tried to find some hack for this but didn't find anything fully working yet. Should I continue to investigate or we will do it in the next sprint?

eleu:
> there is no direct way to do it. I tried to find some hack for this but didn't find anything fully working yet. Should I continue to investigate or we will do it in the next sprint?

OK no problem, in that case let's keep it like that!

src/cesiumutils.js
```diff
@@ -369,3 +382,23 @@ export function getOrthogonalViewPoints(viewer) {
...
+  const rightCartographic = Cartographic.fromCartesian(right);
+  const leftUp = Cartesian3.fromRadians(rightCartographic.longitude, leftCartographic.latitude);
+  const rightDown = Cartesian3.fromRadians(rightCartographic.longitude, rightCartographic.latitude);
+  const isEast = rightCartographic.longitude > leftCartographic.longitude;
+  const isNorth = rightCartographic.latitude > leftCartographic.latitude;
+  return new Cartesian3(
+    Cartesian3.distance(left, leftUp) * (isEast ? 1 : -1),
+    Cartesian3.distance(leftUp, rightDown) * (isNorth ? 1 : -1),
+    rightCartographic.height - leftCartographic.height
+  );
```
gberaudo, line 403: @vladyslav-tk , can you detail what this function does?

src/elements/ngm-nav-tools.ts
```diff
@@ -67,12 +94,95 @@ export class NgmNavTools extends LitElementI18n {
...
+      this.removeReference();
+    } else {
+      this.eventHandler.setInputAction(debounce(event => this.onMoveRef(event), 250), ScreenSpaceEventType.MOUSE_MOVE);
+      this.eventHandler.setInputAction(event => this.onLeftDown(event), ScreenSpaceEventType.LEFT_DOWN);
+      this.eventHandler.setInputAction(() => this.onLeftUp(), ScreenSpaceEventType.LEFT_UP);
+      this.setReference();
+    }
+  }
+
+  setReference(position?: Cartesian3) {
```
gberaudo, line 116: Always passing the position parameter would make the function clearer.

src/elements/ngm-nav-tools.ts
```diff
@@ -67,12 +94,95 @@ export class NgmNavTools extends LitElementI18n {
...
+    this.refIcon.viewFrom = <any>eastNorthUp(center, cam.position);
+    if (!position) {
+      const transform = Transforms.eastNorthUpToFixedFrame(center);
+      cam.lookAtTransform(transform);
+    }
+    document.addEventListener('keydown', this.ctrlListener);
+
+  }
+
+  removeReference() {
```
gberaudo, line 135: What about using the same wording as the user interface? Here and in the other places.
```suggestion
  removeTargetPoint() {
```

src/utils.ts
```diff
@@ -168,3 +168,14 @@ export async function zoomTo(viewer: Viewer, config): Promise<void> {
...
   }
 }
+
+export function debounce(f, ms) {
+  let isCooldown = false;
+  return (...args) => {
+    if (isCooldown) return;
+    f(...args);
+    isCooldown = true;
+    setTimeout(() => isCooldown = false, ms);
```
gberaudo, line 178: Nice to have: ideally the function would also be called at the end of the timeout because currently, if a call is triggered 250ms before the mouse hovers the button the cursor will not change styling.

src/elements/ngm-nav-tools.ts
```diff
@@ -67,12 +94,95 @@ export class NgmNavTools extends LitElementI18n {
...
+    this.moveRef = false;
+    this.viewer!.trackedEntity = this.refIcon;
+    this.viewer!.scene.screenSpaceCameraController.enableInputs = true;
+    // for better performance
+    this.eventHandler!.setInputAction(debounce(event => this.onMoveRef(event), 250), ScreenSpaceEventType.MOUSE_MOVE);
+  }
+
+  onMoveRef(event) {
+    if (this.moveRef) {
+      const position = Cartesian3.clone(this.viewer!.scene.pickPosition(event.endPosition));
```
gberaudo, line 171: Is cloning necessary? For performance you can pass a scratch cartesian to `pickPosition`.

src/elements/ngm-nav-tools.ts
```diff
@@ -67,12 +94,95 @@ export class NgmNavTools extends LitElementI18n {
...
+  onLeftUp() {
+    if (!this.moveRef) return;
+    this.moveRef = false;
+    this.viewer!.trackedEntity = this.refIcon;
+    this.viewer!.scene.screenSpaceCameraController.enableInputs = true;
+    // for better performance
+    this.eventHandler!.setInputAction(debounce(event => this.onMoveRef(event), 250), ScreenSpaceEventType.MOUSE_MOVE);
+  }
+
+  onMoveRef(event) {
```
gberaudo, line 169: This should be called 
```suggestion
  onMouseMove(event) {
```

src/viewer.ts
```diff
@@ -260,6 +259,9 @@ function enableCenterOfRotate(viewer: Viewer) {
   eventHandler.setInputAction(() => {
     scene.camera.lookAtTransform(Matrix4.IDENTITY);
   }, ScreenSpaceEventType.LEFT_UP, KeyboardEventModifier.CTRL);
+  document.addEventListener('keyup', (evt) => {
+    if (evt.key === 'Control') scene.camera.lookAtTransform(Matrix4.IDENTITY);
```
gberaudo, line 263: Could you document why we do this?

src/style/ngm-object-information.css
```diff
@@ -93,3 +93,12 @@ ngm-object-information .content-container {
   overflow-x: hidden;
   max-height: 505px;
 }
+
+ngm-object-information div.content-container td,
+ngm-object-information div.content-container span {
+  -webkit-user-select: text;
+  -khtml-user-select: text;
+  -moz-user-select: text;
+  -o-user-select: text;
```
gberaudo, line 102: Do we need all those legacy prefixes?

gberaudo (approved):
I am very pleased by the results: it works very well and really gives confidence in the navigation experience.

vladyslav-tk:
@gberaudo updates according to comments are done

PR #658, opened 2021-11-05
GSNGM-624: Height slider

b4l:
It is somewhat working on Firefox, it seems that chromium does not support vertical input range though.

There is also some weird behavior that the slider thumb is not updated anymore from navigation after the slider was used to adjust the camera height.

Some stuff left to do:
- [x] Snap to terrain
- [x] Take altitude into account on slider input `valueToHeight()`
- [x] Fix bug


vladyslav-tk:
@b4l I fixed issues

PR #664, opened 2021-11-09
Grey out action menu for kml layers

b4l:
Hide/grey out kml upload layer action menu until further ui/ux specification take place

### DIFF
--- a/src/style/layers.css
+++ b/src/style/layers.css
@@ -63,49 +63,6 @@ ngm-layers-item .ui.dropdown .menu .item:hover {
   background-color: #C5F6FA;
 }
 
-/* KML upload */
-ngm-layers-upload button {
-  cursor: pointer;
-  display: flex;
-  align-items: center;
-  letter-spacing: 0.25px;
-  font-weight: bold;
-  color: #0B7285;
-  background-color: #F1F3F5;
-  border: 2px dashed #0B7285;
-  margin-top: 9px;
-  padding-left: 10px;
-  height: 46px;
-  width: 100%;
-}
-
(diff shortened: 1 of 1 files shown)


src/style/layers.css
```diff
@@ -146,3 +121,47 @@ ngm-catalog .ngm-checkbox {
...
+ngm-layers-upload .ngm-upload-icon {
+  pointer-events: none;
+  background-color: #0B7285;
+  margin-top: 0;
+  margin-left: 12px;
+  width: 20px;
+  height: 20px;
+}
+
+ngm-layers-upload .ngm-upload-icon {
```
vladyslav-tk, line 150: duplicated, see line 141

src/style/layers.css
```diff
@@ -146,3 +121,47 @@ ngm-catalog .ngm-checkbox {
...
+  pointer-events: none;
+  background-color: #0B7285;
+  margin-top: 0;
+  margin-left: 12px;
+  width: 20px;
+  height: 20px;
+}
+
+/* FIXME: Grey out action menue until further specification */
+ngm-layers-upload .ui.dropdown {
```
vladyslav-tk, line 160: it is ok to keep it like this but you can also use `ngm-disabled` instead of adding new styles

vladyslav-tk (approved):

src/style/layers.css
b4l: Elisabeth is ok with grey out, so we might just leave it, anyways it will get removed soon.

src/style/layers.css
b4l: removed it by force push

PR #667, opened 2021-11-15
GSNGM-612: Settings panel

b4l:


vladyslav-tk:
@b4l I fixed the error with setup i18n. You got this error because `#langs` was not loaded on the `firstUpdated` of sidebar and you passed `null` in `render` inside `setupI18n`. Language change still works strange, sometimes it works sometimes show tags without translation

vladyslav-tk:
@b4l Also, I somehow lost your changes after merge but seems like (hope) I added everything back

src/elements/ngm-side-bar.ts
```diff
@@ -320,6 +330,11 @@ export class SideBar extends LitElementI18n {
       }
     }
 
+    if (this.querySelector('#langs') && !this.langInited) {
+      setupI18n();
+      this.langInited = true;
+    }
+
```
gberaudo, line 337: I don't think you need this. The component should re-render automatically once i18n is initialized (initialization is already triggered in ngm-app).

src/i18n.js
```diff
@@ -71,6 +71,10 @@ export function setupI18n() {
     document.documentElement.lang = lang;
     localize('[data-i18n]');
   });
+  const templates = SUPPORTED_LANGUAGES.map(lang => html`
+    <div class="item lang-${lang}" @click="${() => i18next.changeLanguage(lang)}">${lang.toUpperCase()}</div>
+  `);
+  render(templates, document.getElementById('langs'));
```
gberaudo, line 77: This should not be done here. See the other comment.

src/elements/ngm-side-bar.ts
```diff
@@ -144,6 +145,15 @@ export class SideBar extends LitElementI18n {
         ${this.activePanel !== 'share' ? '' : html`
           <ngm-share-link></ngm-share-link>`}
       </div>
+      <div .hidden=${this.activePanel !== 'settings'} class="ngm-side-bar-panel">
+        <div class="ngm-panel-header">${i18next.t('lsb_settings')}
+          <div class="ngm-close-icon" @click=${() => this.activePanel = ''}></div>
+        </div>
+        <div class="language-settings">
+          <label>${i18next.t('lsb_langs')}</label>
+          <div id="langs" class="ui horizontal selection list"></div>
```
gberaudo, line 154: Something like this should work:
```suggestion
          <div id="langs" class="ui horizontal selection list">
          ${SUPPORTED_LANGUAGES.map(lang => html`
            <div class="item lang-${lang}" @click="${() => i18next.changeLanguage(lang)}">${lang.toUpperCase()}</div>
           }
  `);
          </div>
```

gberaudo (changes requested):

src/elements/ngm-side-bar.ts
b4l: ok, reverted so previous initialization, seems to work fine.

gberaudo (approved):

src/i18n.js
b4l: addressed it in https://github.com/swissgeol/ngm/pull/667/commits/6f51e81fa90278d807f3623d6175286e9729f4ac

src/elements/ngm-side-bar.ts
b4l: yes, done in https://github.com/swissgeol/ngm/pull/667/commits/6f51e81fa90278d807f3623d6175286e9729f4ac

PR #668, opened 2021-11-15
Geometry floating info box / editing

vladyslav-tk:


gberaudo:
@vladyslav-tk, I should not be able to save with an empty name.

vladyslav-tk:
> @vladyslav-tk, I should not be able to save with an empty name.

oh yes, I forgot about this, Thanks, fixed

src/toolbox/ngm-geometry-drawer.ts
```diff
@@ -112,9 +91,24 @@ export class NgmAreaOfInterestDrawer extends LitElementI18n {
...
       this.addAreaEntity(conf);
     });
-    QueryStore.objectInfo.subscribe(info => {
-      if (!info) {
+    DrawStore.draw.subscribe(draw => {
+      if (draw) {
+        this.draw = draw;
+        this.draw.addEventListener('statechanged', () => this.requestUpdate());
+        this.draw.addEventListener('drawend', (evt) => this.endDrawing_((<CustomEvent>evt).detail));
+        this.draw.addEventListener('drawerror', evt => {
```
gberaudo, line 99: There are several places where you subscribe or add an event listener but I see no unlistening. Is it correct?

gberaudo (approved):

gberaudo:
@vladyslav-tk, I noticed that:
- when I click on the globe a popup is displayed (OK);
- now I draw a point: it does not disappear completly but becomes very slim.

src/toolbox/ngm-geometry-drawer.ts
```diff
@@ -75,34 +60,28 @@ const DEFAULT_AREAS_COUNTER = {
...
   private sectionImageUrl: string | undefined;
-  private editedBackup;
   private areasClickable = false;
   private unlistenEditPostRender: Event.RemoveCallback | undefined;
   private drawGeometries = [
-    {labelTag: 'tbx_add_point_btn_label', type: 'point', icon: 'ngm-point-draw-icon'},
-    {labelTag: 'tbx_add_line_btn_label', type: 'line', icon: 'ngm-line-draw-icon'},
-    {labelTag: 'tbx_add_polygon_area_btn_label', type: 'polygon', icon: 'ngm-polygon-draw-icon'},
-    {labelTag: 'tbx_add_rect_area_btn_label', type: 'rectangle', icon: 'ngm-rectangle-draw-icon'},
+    {label: i18next.t('tbx_add_point_btn_label'), type: 'point', icon: 'ngm-point-draw-icon'},
```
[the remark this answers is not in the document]
vladyslav-tk: in this case, it works without function (probably renders when translation ready) but I add function anyway to be sure

src/toolbox/ngm-geometry-drawer.ts
vladyslav-tk: we don't need to unlistening these events as we have only one example of geometry-drawer element and we should always listen for draw events

vladyslav-tk:
> @vladyslav-tk, I noticed that:
> 
> * when I click on the globe a popup is displayed (OK);
> * now I draw a point: it does not disappear completly but becomes very slim.

Do you mean object info popup? We have the same bug on the dev, I can have a look and fix it if it is fast or we can create a bug issue for this

vladyslav-tk:
> > @vladyslav-tk, I noticed that:
> > 
> > * when I click on the globe a popup is displayed (OK);
> > * now I draw a point: it does not disappear completly but becomes very slim.
> 
> Do you mean object info popup? We have the same bug on the dev, I can have a look and fix it if it is fast or we can create a bug issue for this

fixed

PR #672, opened 2021-11-18
Dashboard: swisstopo showcase projects (permalinks collection, no project geometries)

vladyslav-tk:


eleu:
really great to see that!

Two small comments:
* Font size of "Description" needs adaption
* Can we activate the target point for some of the links? (maybe means adding it to the permalink, actually)

src/constants.ts
```diff
@@ -105,3 +105,54 @@ export const SHORTLINK_HOST_BY_PAGE_HOST = {
   'test.viewer.swissgeol.ch': 'link.swissgeol.ch',
   'viewer.swissgeol.ch': 'link.swissgeol.ch',
 };
+
+export const SHOWCASE_PROJECTS = [
```
gberaudo, line 109: I thinks this should be put in a static file and fetched when the dashboard is opened for the first time.

gberaudo (approved):
LGTM

src/constants.ts
vladyslav-tk: updated

PR #673, opened 2021-11-18
do not overlap floating windows on open

vladyslav-tk:


src/elements/draggable.ts
```diff
@@ -0,0 +1,85 @@
...
+  });
+  target.interaction && target.interaction.reflow({name: 'drag', axis: 'xy'});
+}
+
+function repositionOnOpen(target) {
+  const observer = new MutationObserver(((mutationsList) => {
+    if (mutationsList[0].attributeName === 'hidden' && !target.hidden) {
+      let overlapList: DOMRect[] = checkForOverlap(target);
+      let counter = 0;
+      while (overlapList.length !== 0) {
```
gberaudo, line 74: While loops are the most evil of all constructs. I propse a soft for i < 10 and break when the overlap list is empty

src/elements/draggable.ts
```diff
@@ -0,0 +1,85 @@
...
+      const targetRect = target.getBoundingClientRect();
+      const checkRect = elem.getBoundingClientRect();
+      const overlap = !(targetRect.right < checkRect.left || targetRect.left > checkRect.right);
+      if (overlap) overlapList.push(checkRect);
+    }
+  });
+  return overlapList;
+}
+
+function moveWindow(target, overlapList, changeSide) {
```
gberaudo, line 48: I don't understand all the logics here. Could you add some comments?

I would expect to simply try positioning at top: 0, right: XX . Then if there is collision move the target to the left of the colliding window, check collision and move again if needed. In the case where we are too far on the left (window would go out of the the viewport), then I would put the window at right: XX.
XX: the left + margin of the +0- bar.
We should also not overlap the panels. If there is an overlap => top: 0, right: XX.

That said I am not against a more clever logics.

vladyslav-tk: I added comments.

As I said it works not perfectly now but I already spent 4h that issue was estimated (+ it is an investigation issue) so I am not sure that I should continue making updates

gberaudo (approved):

PR #675, opened 2021-11-19
GSNGM-644: Geometry filters (+ GSNGM-661)

vladyslav-tk:


eleu:
great! Can we make the background color of this button group normal-panel-grey (#f1f3f5) and not the very-light grey? (while keeping the very light grey for the geometries)

vladyslav-tk:
> great! Can we make the background color of this button group normal-panel-grey (#f1f3f5) and not the very-light grey? (while keeping the very light grey for the geometries)

fixed

fredj (approved):

PR #678, opened 2021-11-22
Use 5 digits for WGS84 coordinates

gberaudo:


fredj (approved):

fredj:
I pushed a commit to update the value displayed in the storybook

PR #682, opened 2021-11-22
Allow height slider to go to 300000 + slider colors

fredj:


eleu:
@fredj can we set to 0 value in the middle of the slider? (so, the left side spanning over 90 px and -30 to 0 km, and the right side also 90 px wide and 0 to 300 km)

fredj:
> @fredj can we set to 0 value in the middle of the slider? (so, the left side spanning over 90 px and -30 to 0 km, and the right side also 90 px wide and 0 to 300 km)

So the 'zero' of the slider will be at 165'000m ?

eleu:
@fredj No, the 0 is in the middle. 1 px will be 333 meters on the left side, and 3'333 meters on the right side.

src/elements/ngm-cam-configuration.ts
```diff
@@ -23,6 +23,19 @@ import './ngm-cam-coordinates';
...
+  return m * height + 0.5;
+}
+
+function valueToHeight(value: number): number {
+  if (value < 0.5) {
+    return (30000 / 0.5) * value - 30000;
+  } else {
+    return (300000 / 0.5) * value - 300000;
+  }
+}
```
gberaudo, line 37: These formulas look esoteric to me. Could you add a comment about what they do?
Concretly, I suppose height is the cartographic height.
But what is value? And its units?

fredj: comments added

src/elements/ngm-cam-configuration.ts
```diff
@@ -23,6 +23,29 @@ import './ngm-cam-coordinates';
 
 type LockType = '' | 'elevation' | 'angle' | 'pitch' | 'move';
 
+/*
+ * Convert cartographic height (between -30'000m and +300'000) to input value (between 0 and 1)
+ * The input value between 0 to 0.5 is mapped to the height between -30'000m and 0m
+ * The input between 0.5 and 1 is mapped to the height between 0m and +300'000m
```
gberaudo, line 29: Thanks a lot @fredj , that is super clear now!

src/elements/ngm-cam-configuration.ts
```diff
@@ -23,6 +23,29 @@ import './ngm-cam-coordinates';
...
+  const m = 0.5 / (height < 0 ? 30000 : 300000);
+  return m * height + 0.5;
+}
+
+/*
+ * Convert input value (between 0 and 1) to cartographic height (between -30'000m and +300'000)
+ * The input value between 0 to 0.5 is mapped to the height between -30'000m and 0m
+ * The input between 0.5 and 1 is mapped to the height between 0m and +300'000m
+ */
+function valueToHeight(value: number): number {
```
gberaudo, line 41: What about creating unit tests for these 2 functions? Would be straightforward and a strong guarantee everything works as expected. Just export them and add a file to src/tests.

src/elements/ngm-height-slider.ts
```diff
@@ -46,19 +46,31 @@ export class NgmHeightSlider extends LitElementI18n {
...
+  /*
+   * Convert input value (between 0 and 3) to cartographic height (between -30'000m and +300'000)
+   * The input value between 0 and 1.35 is mapped to the height between -30'000m and 0m
+   * The input value between 1.35 and 2.7 is mapped to the height between 0m and +30'000m
+   * The input value between 2.7 and 3 is mapped to the height between +30'000m and +300'000m
+   */
   valueToHeight(value: number) {
-    if (value >= 2) {
-      return 300000 * (value - 2);
+    if (value < 2.7) {
```
gberaudo, line 70: Idem for these

gberaudo (approved):
LGTM, please add UT if you want.

eleu:
(OK also from the PO side. I explained the special slider situation in the Jira issue. I let you see for the UT.)

PR #688, opened 2021-11-25
GSNGM-708: Fix zoom to geometries

vladyslav-tk:


src/toolbox/ngm-geometry-drawer.ts
```diff
@@ -242,16 +240,15 @@ export class NgmAreaOfInterestDrawer extends LitElementI18n {
     this.draw!.active = true;
   }
 
-  flyToArea(id) {
+  async flyToArea(id) {
```
gberaudo, line 243: Do you need this async?

gberaudo (approved):

src/toolbox/ngm-geometry-drawer.ts
vladyslav-tk: forgot to remove, thanks

PR #698, opened 2021-12-01
Height slider: snap to 10m above terrain when camera is moved to the terrain limit

vladyslav-tk:


src/elements/ngm-height-slider.ts
```diff
@@ -70,7 +70,7 @@ export class NgmHeightSlider extends LitElementI18n {
     const altitude = this.viewer.scene.globe.getHeight(camera.positionCartographic) || 0;
     let height = Math.round(valueToHeight(Number(input.value)));
     const snapDistance = 150; // snap to 0 from 150m/-150m
```
eleu, line 72: Can you also change the snapDistance to f.e. 800? (currently I nearly don't feel that it snaps, I'd expect that it would snap as soon as the icon approaches the limit. This is at about +/-1000 meters, but it depends of course on the screen size. Can we try 800m therefore?)

eleu (approved):

eleu:
the 10m limit looks good, thanks! We should just make it a bit more "magnetic", cf. my comment above.

src/elements/ngm-height-slider.ts
vladyslav-tk: updated

PR #709, opened 2021-12-07
GSNGM-716: Download data in an area: add the button to geom info panel

b4l:


eleu:
I just had a very quick look, not sure but maybe this is missing?

> Do not show the button if no data can be downloaded (f.e. if there is no layer with downloadable data switched on).

eleu:
(maybe it's just that not all cases have been described in the issue, as usual ;) The one which seems to miss is: Download of data is only possible for rectangles at the moment, and if a "data downloadable" layer is visible in 3D viewer (boreholes, cross sections, maybe others...).

fredj (approved):

PR #722, opened 2021-12-15
Slicing: Zoom to the geometry when slicing is switched on for a line / rectangle

vladyslav-tk:


b4l (approved):
Works as advertised! Was wondering if we should extract a generic function for zooming to an arbitrary box. Seems like a common use case.

vladyslav-tk:
> Works as advertised! Was wondering if we should extract a generic function for zooming to an arbitrary box. Seems like a common use case.

yes agree, moved it to helpers

PR #727, opened 2021-12-16
Allow to tweak light

gberaudo:


src/viewer.ts
```diff
@@ -181,8 +166,25 @@ export function setupViewer(container: Element, rethrowRenderErrors) {
   }
 
   // Position the sun the that shadows look nice
-  viewer.clock.currentTime = JulianDate.fromDate(new Date('June 21, 2018 12:00:00 GMT+0200'));
-
+  let sunDate = new Date('2018-06-21T10:00:00.000Z');
+  if (searchParams.has('date')) {
+    const betterDate = new Date(searchParams.get('date') || 'qwer');
```
fredj, line 171: Can we remove 'qwer' and let the date be invalid by the content of the param?

fredj (approved):

src/viewer.ts
gberaudo: `qwer` is for tslint because searchParams.get can return `null`. It looks like Date does not accept null as parameter. But I can use an empty string for the same result ;)

PR #739, opened 2021-12-20
Show snackbar regarding the z / angle / pitch / coordinates fixation when a user closes the camera panel

vladyslav-tk:
…closes the camera panel

### DIFF
--- a/src/elements/ngm-cam-configuration.ts
+++ b/src/elements/ngm-cam-configuration.ts
@@ -19,6 +19,7 @@ import {styleMap} from 'lit/directives/style-map.js';
 import {classMap} from 'lit/directives/class-map.js';
 import './ngm-cam-coordinates';
 import NavToolsStore from '../store/navTools';
+import {showSnackbarInfo} from '../notifications';
 
 export type LockType = '' | 'elevation' | 'angle' | 'pitch' | 'move';
 
@@ -103,6 +104,16 @@ export class NgmCamConfiguration extends LitElementI18n {
       lock: () => this.toggleLock('pitch')
     },
   ];
+
+  constructor() {
+    super();
+    const observer = new MutationObserver(((mutationsList) => {
+      if (mutationsList[0].attributeName === 'hidden' && !this.hidden) {
+        (<HTMLElement>document.querySelector('.ngm-cam-lock-info'))?.parentElement?.remove();
+      }
+    }));
+    observer.observe(this, {attributes: true});
+  }
 
   connectedCallback() {
     draggable(this, {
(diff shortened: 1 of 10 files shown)


eleu:
works well, thanks! Note: I renamed the issue in Jira because the title was a bit misleading. We do not de-activate it directly, but mainly show a snackbar when closing the camera config... And this is implemented as planned, thanks!

src/elements/ngm-cam-configuration.ts
```diff
@@ -253,12 +264,35 @@ export class NgmCamConfiguration extends LitElementI18n {
     }
   }
 
+  onClose() {
+    this.dispatchEvent(new CustomEvent('close'));
+    if (this.lockType) {
+      // keep commented tags to have translations
+      // i18next.t('cam_lock_info_elevation') i18next.t('cam_lock_info_angle')
+      // i18next.t('cam_lock_info_pitch') i18next.t('cam_lock_info_move')
+      showSnackbarInfo(i18next.t(`cam_lock_info_${this.lockType}`),
```
gberaudo, line 273: Instead of relying on comments I would rather have an explicit switch:
```js
switch (this.lockType):
case 'angle':
  tr = i18next.t('cam_lock_info_angle');
  break;
...
```

gberaudo: Do we want translation to change when the language change? If so we should let  `showSnackbarInfo` call i18next.t.

src/elements/ngm-cam-configuration.ts
```diff
@@ -253,12 +264,35 @@ export class NgmCamConfiguration extends LitElementI18n {
...
+        {
+          displayTime: 0,
```
gberaudo, line 275: If the info should be displayed `forever`, I would rather pass `-1`, or eventually not `Number.MAX_VALUE`.

src/elements/ngm-cam-configuration.ts
```diff
@@ -103,6 +104,16 @@ export class NgmCamConfiguration extends LitElementI18n {
     },
   ];
 
+  constructor() {
+    super();
+    const observer = new MutationObserver(((mutationsList) => {
+      if (mutationsList[0].attributeName === 'hidden' && !this.hidden) {
```
gberaudo, line 110: Why do we need an observer to be notified of a change of a property we control?

src/elements/ngm-cam-configuration.ts
vladyslav-tk: according to https://fomantic-ui.com/modules/toast.html#/settings it should be 0

src/elements/ngm-cam-configuration.ts
vladyslav-tk: replaced with a switch. 

To call i18next from `showSnackbarInfo` and make it work we should have some big switch with all messages, not sure that it a good way

src/elements/ngm-cam-configuration.ts
vladyslav-tk: updated

gberaudo (approved):

PR #742, opened 2021-12-21
Avoid de-centering or out of view of the target point

vladyslav-tk:


eleu:
there seems to be two small details missing: we should also de-activate the target point when the user uses the the Angle / Pitch **sliders**.

vladyslav-tk:
> there seems to be a small detail missing: we should also de-activate the target point when the user uses the the Angle / Pitch **sliders**.

fixed

eleu:
great thanks, works well!

fredj (approved):

PR #753, opened 2021-12-23
Multidrag

gberaudo:


gberaudo:
@fredj, the storybook I added looks like working great, but not yet the app (which does nothing).
That is a bit puzzling. Maybe it is something with the handler, I will investigate.

PR #761, opened 2022-01-10
2021.18 dashboard

anjaamrein:


src/sampleData/showcase_projects.json
```diff
@@ -40,5 +40,26 @@
...
+  {
+    "title": "Water Tower Switzerland",
+    "description": "Hydropower is responsible for generating around 57 % of Switzerland’s electricity. Like veins, countless water tunnels and conduit shafts criss-cross the Alps, channelling the precious resource with the aim of producing energy.",
+    "created": "10 am January 2022",
+    "image": "./images/dam_LacDesDix.png",
+    "color": "#B9271A",
+    "views": [
+      {
+        "title": "Largest storage hydropower storage plant in Bieudron with 2271 GWh per year",
+        "permalink": "?lon=7.21753&lat=46.22802&elevation=3516&heading=156&pitch=-22&map=ch.swisstopo.swissimage&map_transparency=0.41&layers=ch.swisstopo.swisstlm3d.3d%2Cwater_tunnel%2Cch.swisstopo.swissnames3d.3d%2Cch.swisstopo.geologie-geocover&layers_visibility=true%2Ctrue%2Ctrue%2Cfalse&layers_transparency=0.00%2C0.00%2C0.00%2C1.00&target=%7B%22lon%22%3A%228.60100%22%2C%22lat%22%3A%2246.70688%22%2C%22height%22%3A932.9088491526485%7D&lang=en"
```
eleu, line 53: the target point is out of view with this link, when I test it with the demo. Maybe remove the target point or set it at a in-view place?

src/sampleData/showcase_projects.json
```diff
@@ -40,5 +40,26 @@
...
+      },
+      {
+        "title": "Geology of the Rhone Valley",
+        "permalink": "?lon=7.20907&lat=46.24121&elevation=4165&heading=156&pitch=-22&map=ch.swisstopo.swissimage&map_transparency=0.41&layers=ch.swisstopo.swisstlm3d.3d%2Cwater_tunnel%2Cch.swisstopo.swissnames3d.3d%2Cch.swisstopo.geologie-geocover&layers_visibility=true%2Ctrue%2Ctrue%2Ctrue&layers_transparency=0.00%2C0.00%2C0.00%2C0.51&target=%7B%22lon%22%3A%228.60100%22%2C%22lat%22%3A%2246.70688%22%2C%22height%22%3A932.9088491520051%7D&lang=en"
```
eleu, line 57: same problem here with the target point.

src/sampleData/showcase_projects.json
```diff
@@ -40,5 +40,26 @@
...
+      },
+      {
+        "title": "Natural hazards: Can earthquakes be a threat?",
+        "permalink": "https://int.swissgeol.ch/?lon=7.21753&lat=46.22802&elevation=3516&heading=156&pitch=-22&map=ch.swisstopo.swissimage&map_transparency=0.76&layers=ch.swisstopo.swisstlm3d.3d%2Cwater_tunnel%2Cch.swisstopo.swissnames3d.3d%2Cearthquakes%2Chistorical_earthquakes%2Cch.swisstopo.geologie-geocover&layers_visibility=true%2Ctrue%2Ctrue%2Ctrue%2Ctrue%2Ctrue&layers_transparency=0.00%2C0.00%2C0.00%2C0.00%2C0.00%2C0.97&target=%7B%22lon%22%3A%228.60100%22%2C%22lat%22%3A%2246.70688%22%2C%22height%22%3A932.9088491520051%7D&lang=en"
```
eleu, line 61: same problem here with the target point

src/sampleData/showcase_projects.json
```diff
@@ -40,5 +40,26 @@
         "permalink": "?lon=8.59559&lat=46.69875&elevation=417&heading=21&pitch=23&map=ch.swisstopo.swissimage&map_transparency=0.43&layers=ch.swisstopo.swisstlm3d.3d,ch.swisstopo.geologie-geocover,water_tunnel,rail_tunnel,road_tunnel&layers_visibility=true,false,true,true,true&layers_transparency=0.00,0.82,0.00,0.00,0.000&target={\"lon\":\"8.60072\",\"lat\":\"46.70794\",\"height\":881.8231693363072}"
       }
     ]
+  },
+  {
+    "title": "Water Tower Switzerland",
+    "description": "Hydropower is responsible for generating around 57 % of Switzerland’s electricity. Like veins, countless water tunnels and conduit shafts criss-cross the Alps, channelling the precious resource with the aim of producing energy.",
+    "created": "10 am January 2022",
```
eleu, line 47: Can you correct that to "10 January 2022"? It's wrong also in the previous projects... It was a confusion between date and time ("am" / "pm") and German ("Erstellt am...").

If it is easy, change it also for the other old projects listed above!

The "created" is currently just a text, you don't have to follow a specific format. (maybe we have to change it one day, but let's handle it like that for the moment).

src/sampleData/showcase_projects.json
anjaamrein: I did not include the target point on purpose, I removed it for all those 3 views. 
Transparencies are fixed in the permalink as intended for the different views. But for the layer GeoCover and the Background map, opacity are not displayed as defined in the permalink. When I operate and I slightly change opacity in the menu operating button, the right opacity is displayed. Without any changes, opacity of those two layers is displayed as 100% instead of 59% for example.

src/sampleData/showcase_projects.json
anjaamrein: Target point removed

src/sampleData/showcase_projects.json
anjaamrein: Target point removed.

src/sampleData/showcase_projects.json
anjaamrein: Time changed. I also did not get the point about the "am". I removed it also for the other dashboard topics.

PR #770, opened 2022-01-18
Add back ngm-feature-height

fredj:


gberaudo (approved):
What about terrain height? Should it be displayed when there is no object intersecting? The spec is not clear to me.

PR #781, opened 2022-01-21
GSNGM-758: Improvements

vladyslav-tk:


gberaudo (commented):
Could you describe what the the code is doing?

eleu:
@gberaudo probably these (it was my testing feedback) :

- When drawing a geom and the geom list is very long, scroll down so the user sees the open geom
- When drawing f.e. a rectangle and the user has filtered for lines, remove the filter (= show all) so the rectangle is visible in the list (maybe this is a general thing to do for all geom's lists in the Tools).

=> just tested, looks good, thanks!

eleu (approved):

PR #786, opened 2022-01-24
GSNGM-670: Differentiate the layer if they are 3D in "Data displayed"

b4l:


src/layers/ngm-layers-item.ts
```diff
@@ -75,6 +76,16 @@ export class LayerTreeItem extends LitElementI18n {
...
     this.actions.changeVisibility(this.config, this.config.visible);
     this.dispatchEvent(new CustomEvent('layerChanged'));
+    // display visibility hint for subsurface layers
+    if (this.config.visible
+      && [LayerType.tiles3d, LayerType.earthquakes].includes(this.config.type!)) {
+      const sidebar = <any> document.getElementsByTagName('ngm-side-bar')[0];
+      if (sidebar.displayUndergroundHint
+        && !sidebar.viewer?.scene.cameraUnderground) {
+        showSnackbarInfo(i18next.t('lyr_subsurface_hint'));
+        sidebar.displayUndergroundHint = false;
```
vladyslav-tk, line 86: better to replace it with customEvent or use rxjs or modify and use `layerChanged` event (it handles in side-bar) instead of directly getting the element

src/elements/ngm-side-bar.ts
```diff
@@ -406,6 +407,13 @@ export class SideBar extends LitElementI18n {
       layer.visible = true;
       layer.displayed = true;
       this.activeLayers.push(layer);
+      // display visibility hint for subsurface layers
+      if (this.displayUndergroundHint
+        && [LayerType.tiles3d, LayerType.earthquakes].includes(layer.type!)
+        && !this.viewer?.scene.cameraUnderground) {
+        showSnackbarInfo(i18next.t('lyr_subsurface_hint'));
```
vladyslav-tk, line 414: according to the description, display time should be the same as in the navigation hint (20 sec)

b4l: addressed in https://github.com/swissgeol/ngm/pull/786/commits/db524154538145c673aa40074b9879a8f272fedd

src/layers/ngm-layers-item.ts
b4l: used custom event in https://github.com/swissgeol/ngm/pull/786/commits/db524154538145c673aa40074b9879a8f272fedd

vladyslav-tk (approved):

PR #787, opened 2022-01-25
GSNGM-780: Show / hide layers does not work (eye icon)

b4l:


vladyslav-tk (approved):

b4l:
@vladyslav-tk do you know the reason why it needs a manual trigger? after all it is a property.

vladyslav-tk:
> @vladyslav-tk do you know the reason why it needs a manual trigger? after all it is a property.

yes, `@property` or `@state` trigger update only when you reassign it. For example: 

* `this.config.opacity = Number(input.value);` not trigger update
* `this.config = config;` trigger update

PR #795, opened 2022-01-28
GSNGM-756: Change the shortlink service to use Database

b4l:


eleu:
wow great it seems to work and it is super fast now!

fredj (approved):

PR #802, opened 2022-02-03
Profile: show place of the profile on the map

vladyslav-tk:


src/cesiumutils.ts
```diff
@@ -284,6 +284,34 @@ export function projectPointOnSegment(point, startPoint, endPoint, start, end, h
...
 }
 
+/**
+ * Gets distance to point in percentage and returns cartesian position on the line
+ *
+ * @param linePositions
+ * @param distance - distance to point in percentage (value from 0 to 1 where 0 first point of the line and 1 is last)
+ * @param result
+ */
+export function getPointOnPolylineByPercentage(linePositions: Cartesian3[], distance: number, result = new Cartesian3()) {
```
gberaudo, line 294: "distance" shoudl be called "ratio".

src/cesiumutils.ts
```diff
@@ -284,6 +284,34 @@ export function projectPointOnSegment(point, startPoint, endPoint, start, end, h
...
+  let prevRatio, currRatio, newRatio, indx = 0;
+  for (let i = 0; i <= totalPoints; i++) {
+    if (i > 0) prevRatio = currRatio;
+    currRatio = i / totalPoints;
+    if (distance > prevRatio && distance <= currRatio) {
+      indx = i;
+      newRatio = ((distance - prevRatio) / currRatio) * i;
+      break;
+    }
+  }
```
gberaudo, line 305: This block should be replaced with simple maths (no need to iterate).
What is newRatio?

src/cesiumutils.ts
```diff
@@ -284,6 +284,34 @@ export function projectPointOnSegment(point, startPoint, endPoint, start, end, h
...
+  if (indx > 0) {
```
gberaudo, line 306: Why 0 is not handled? Should return *linePositions[0]*.

src/cesiumutils.ts
vladyslav-tk: How we can replace iteration with simple math?

newRatio is a ratio of segment (renamed)

gberaudo: I mean something looking like (not tested):

```
position = ratio * totalPoints
prevRatio = Math.floor(position) / totalPoints
currRatio = Math.ceil(position) / totalPoints
segmentRatio = (((ratio - prevRatio) / currRatio) * Math.floor(position)
```

src/cesiumutils.ts
```diff
@@ -284,6 +284,33 @@ export function projectPointOnSegment(point, startPoint, endPoint, start, end, h
...
 }
 
+/**
+ * Gets distance to point in percentage and returns cartesian position on the line
+ *
+ * @param linePositions
+ * @param ratio - distance to point in percentage (value from 0 to 1 where 0 first point of the line and 1 is last)
+ * @param result
+ */
+export function getPointOnPolylineByRatio(linePositions: Cartesian3[], ratio: number, result = new Cartesian3()) {
```
gberaudo, line 294:
```suggestion
export function getPointOnPolylineByRatio(linePositions: Cartesian3[], ratio: number, result)) {
```
Let's require the result parameter since not passing it does not work anyway.

gberaudo:
LGTM now @vladyslav-tk (except the typo `pervDist`).

PR #804, opened 2022-02-04
Limit query result floating panel width

fredj:


src/style/ngm-object-information.css
```diff
@@ -2,6 +2,7 @@ ngm-object-information {
   position: absolute;
   z-index: 2;
   width: 528px;
+  max-width: 100%;
   max-height: 643px;
```
vladyslav-tk, line 6: need to adapt height too

image

fredj: fixed, can you check again ?

vladyslav-tk (approved):

PR #807, opened 2022-02-07
GSNGM-789: Topics with KML models

b4l:


src/elements/ngm-dashboard.ts
```diff
@@ -42,13 +51,26 @@ export class NgmDashboard extends LitElementI18n {
...
     });
     MainStore.viewer.subscribe(viewer => this.viewer = viewer);
+    MainStore.layersRemoved.subscribe(async () => {
+      if (this.selectedViewIndx !== undefined && this.assets) {
+        await Promise.all(this.assets.map(async layer => {
+          const data = await layer.promise;
+          data.show = true;
+          await (<any>document.getElementsByTagName('ngm-side-bar')[0]).onCatalogLayerClicked({
+            detail: {layer}
+          });
```
gberaudo, line 70: It does not look good to simulate click on something of a distance component.
Is there a cleaner way to do it?

fredj: The function is badly named because it's not a click but a custom event.  We have the exact same code here:
https://github.com/swissgeol/ngm/blob/master/src/layers/ngm-layers-upload.ts#L56-L60

This to use events ?

fredj: What about this: https://github.com/swissgeol/ngm/commit/ed58de26c0d04ef39d3c59fed27fc86ad13b0df6 ?
Use CustomEvent instead of function call. The event name should be changed to something else (any name proposition ?)

vladyslav-tk: I think it is better to use rxjs here. I thought about it yesterday when made my updates. I can update this too

vladyslav-tk: oh, I didn't notice your commit, Fred. CustomEvent looks good too but maybe it should be layeradd instead of click

vladyslav-tk: but in this case, we lost await, and probably layers will be added incorrectly. Need to test

vladyslav-tk: It works fine

fredj: I've pushed my commit to this branch
@vladyslav-tk: I also think rxjs is the way to go

vladyslav-tk (approved):

eleu:
@b4l  can we, just to make it testable, add two KML's to the topic "Hot springs"? (I'll upload samples to the Jira issue)

src/elements/ngm-dashboard.ts
gberaudo: Thanks for the improvement. Looks more "webcomponent" now.

eleu:
can we also rename the file "showcase-projects" to topics? just to make it clear also for the future. (I know this is maybe too general for this PR... maybe it needs another PR.)

PR #816, opened 2022-02-10
Share link for a topic and share a topic view link

vladyslav-tk:


eleu:
great, generally works nice!

some changes are needed, maybe the specs were not so clear...:
* view-link: the share icon should not open a context menu. it should just directly copy the view link to the clipboard (and show the success message that it has been copied.
* When the user opens a view-link shared URL, the dashboard topic panel shows quickly and closes directly again. Can we avoid that somehow, meaning that the topic panel does not show up? (but the user would see directly the mini-panel and thus also the active view-link)?
* Maybe a small bug: when the user opens a topic link, f.e. https://review.swissgeol.ch/GSNGM-791/index.html?topicId=3, the view link share button and also the topic context menu does not open any more.

eleu:
great, thanks! also very happy to see the showcase projects renamed to topics.

PR #827, opened 2022-02-16
Simplify & improve docker setup

b4l:
This prevents rebuilding between runs and facilitates using sqlx macros.

Format & cllippy still have to get integrated in CI. Also I propose to integrate the docker builds similar to `abbreviator` so that we can get rid of the `Make` file and a few scripts.

cc @fredj @vladyslav-tk


docker-compose.yaml
```diff
@@ -68,7 +71,7 @@ services:
       POSTGRES_PASSWORD: *db_pass
       POSTGRES_DB: *db_name
     ports:
-      - 15432:5432
+      - 5432:5432
```
gberaudo, line 74: We need a different port because some postgresql may already run on the dev machine.

docker-compose.yaml
```diff
@@ -1,27 +1,30 @@
...
       APP_PORT: "3000"
-      SQLX_OFFLINE: "true"
-    command: ["cargo", "watch", "--shell", "cargo run --offline --target x86_64-unknown-linux-musl"]
+      DATABASE_URL: "postgres://www-data:www-data@db:5432/swissgeol-local"
     volumes:
       - ./api/src:/app/src:ro
-      - ./api/scripts:/app/scripts:ro
       - ./api/migrations:/app/migrations:ro
+      - ./api/Cargo.toml:/app/Cargo.toml:ro
+    command: cargo watch -x "run --target x86_64-unknown-linux-musl"
```
gberaudo, line 27: I prefer the "--shell" version which allows the offline mode. Being in offline mode prevents unneeded crates index update.

docker-compose.yaml
```diff
@@ -1,27 +1,30 @@
...
-      ENVIRONMENT_NAME: local-dev
       PGUSER: &db_user www-data
       PGPASSWORD: &db_pass www-data
       PGDATABASE: &db_name swissgeol-local
       PGHOST: db
       PGPORT: "5432"
       APP_PORT: "3000"
-      SQLX_OFFLINE: "true"
-    command: ["cargo", "watch", "--shell", "cargo run --offline --target x86_64-unknown-linux-musl"]
+      DATABASE_URL: "postgres://www-data:www-data@db:5432/swissgeol-local"
```
gberaudo, line 22: Is there a way to not duplicate the db infos?

api/README.md
```diff
@@ -21,48 +19,39 @@ The latter can be installed with the following command
...
-Check the documentation for further commands.
-
-### sqlx offline
+### `sqlx` offline
 
-To use sqlx compile time verification without acces to a running database use it's `offline` mode.
+To use sqlx compile time verification without acces to a running database use it's `offline` mode which creates creates & updates the [sqlx-data.json](./sqlx-data.json) file.
 
-Creating and updating the database/query definitions run
+This is only required for building release grade containers and can be kept in sync with with
```
gberaudo, line 53: Some typos here and there.

api/README.md
```diff
@@ -21,48 +19,39 @@ The latter can be installed with the following command
...
-
-Start the application.
+### Quickstart
 
 ```bash
-cargo run
+# Start the docker-compose development composition
+make run
+# Run tests
+carg test
```
gberaudo, line 28: cargo

api/Dockerfile
```diff
@@ -19,22 +25,10 @@ RUN adduser \
...
-
-# First we handle fetching and building our dependencies
-# We need a stub entry point for that
-COPY Cargo.toml Cargo.lock ./
-COPY ./src/stub.rs ./src/main.rs
-RUN cargo fetch --target x86_64-unknown-linux-musl
-RUN cargo build --offline --target x86_64-unknown-linux-musl --release
-
-COPY ./ .
+# Build release
```
gberaudo, line 28: Can you keep the 2 steps build system? It is an important optimization that is allowing us to iterate faster.
Building all dependencies currently takes ~2min and it could go up when we add more. With your change we will have to pay for a full compilation each time we start the composition.
By building the dependencies separately it only takes 5s to build the api.

api/.dockerignore
```diff
@@ -1,4 +1,5 @@
 /target
-/Dockerfile*
+/tests
```
gberaudo, line 2: With this change no tests are run in the CI anymore. Is it really what you want?

docker-compose.yaml
```diff
@@ -1,27 +1,30 @@
...
 services:
   api:
-    image: camptocamp/swissgeol_local_api:latest
-    init: true # handle kill signals for rust
+    build:
+      context: ./api
+      dockerfile: Dockerfile
+      args:
+        - ENV=development
+      target: builder
```
gberaudo, line 9: It feels interesting to reuse the same Dockerfile for release and for local dev.
But, what about:
- the speed: using release mode takes much longer to compile
- can we compile dependencies in release mode but api in debug mode (like you do here)?
- the debugger experience, is it the same in release / debug mode (from my experience of C/C++ release mode have deep optimiztion making the code flow sometime impossible to follow step by step)?

gberaudo (changes requested):

docker-compose.yaml
b4l: Addressed in https://github.com/swissgeol/ngm/pull/827/commits/64d2e491ea8b53a42bd1f8ab13bee9b45c89a357

docker-compose.yaml
b4l: Addressed in https://github.com/swissgeol/ngm/pull/827/commits/64d2e491ea8b53a42bd1f8ab13bee9b45c89a357

api/README.md
b4l: Addressed in https://github.com/swissgeol/ngm/pull/827/commits/64d2e491ea8b53a42bd1f8ab13bee9b45c89a357

api/Dockerfile
b4l: Fetching & building is done in the dev container when the `watch` command is invoked through docker-compose. It should reuse the image with all the cache and facilitate incremental builds between docker-compose runs.

api/.dockerignore
b4l: Added Rust CI in https://github.com/swissgeol/ngm/pull/827/commits/64d2e491ea8b53a42bd1f8ab13bee9b45c89a357

docker-compose.yaml
b4l: Release mode is only used when building all stages of the image, for dev only the first one is built then we call watch which uses debug build. Release & debug builds do not share cache.

vladyslav-tk:
@b4l, as we discussed on the daily, there is some problem with environment variables. cargo-watch does not install when the condition is here and API does not run:

image

api/Dockerfile
gberaudo: Will test when the local dev works again.
Since you removed the `--offline` parameter rust will regularily spend time re-downloading the crates index, right?

api/.env
```diff
@@ -1,12 +1,12 @@
...
-APP_PORT=8000
+APP_PORT=3000
 
 # Database
-PGUSER=postgres
-PGPASSWORD=password
-PGHOST=localhost
-PGPORT=54322
-PGDATABASE=swissgeol
+DB_USER=postgres
```
gberaudo, line 5: Instead of introducing app-specific env variables, I find it convenient to directly use postgres variables. It makes it a breeze to use tools from the postgresql exosystem like psql when interacting with the db.

b4l: Converted in https://github.com/swissgeol/ngm/pull/827/commits/6f41070f1908aa780de1cc8d452df9cb13a4f956

api/Dockerfile
```diff
@@ -1,10 +1,16 @@
-## Builder
+## Setup builder
 FROM rust:latest AS builder
 
 RUN rustup target add x86_64-unknown-linux-musl
-RUN rustup component add clippy rustfmt
 RUN apt-get update && apt-get install -y musl-tools musl-dev
-#RUN update-ca-certificates
+RUN if [ "$ENV" = "development"] ; then cargo install cargo-watch ; fi
```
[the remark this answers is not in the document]
b4l: Fixed in https://github.com/swissgeol/ngm/pull/827/commits/48dc884e1729bdb9d22ff1ac3ff1336d57470abb

Makefile
```diff
@@ -3,23 +3,9 @@ export DOCKER_BASE = camptocamp/swissgeol
...
-	docker-compose up
-
-.PHONY: build_local_api
-build_local_api:
-	docker build -t $(DOCKER_BASE)_local_api:latest --build-arg "GIT_HASH=$(GIT_HASH)" -f api/DockerfileDev api
-
-.PHONY: build_api
-build_api:
-	docker build -t $(DOCKER_BASE)_api:latest --build-arg "GIT_HASH=$(GIT_HASH)" api
+run:
```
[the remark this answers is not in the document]
b4l: It should now work without depending on `docker-compose build`

b4l:
@vladyslav-tk it should be fixed in https://github.com/swissgeol/ngm/pull/827/commits/48dc884e1729bdb9d22ff1ac3ff1336d57470abb, can you test it?

vladyslav-tk:
> @vladyslav-tk it should be fixed in [48dc884](https://github.com/swissgeol/ngm/commit/48dc884e1729bdb9d22ff1ac3ff1336d57470abb), can you test it?

yes, fixed

.github/workflows/rust.yml
```diff
@@ -0,0 +1,101 @@
...
+    tags-ignore:
+      - '**'
+    paths:
+      - 'api/**'
+
+env:
+  CARGO_TERM_COLOR: always
+
+
+jobs:
```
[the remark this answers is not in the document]
b4l: I guess this is kind of a trade of for having a cleaner code base and a performant CI/CD. Most actions are used to checkout the repository install the Rust tool chain and caching it. What is actually run are then just plain commands as you can locally as well.  It is kind of a hypothetical fear, also I would rather rely on something that is used by many opposed too a hand made docker quirk.

Actually, most importantly, what we gain here is that we actually run the tests now which we do not currently.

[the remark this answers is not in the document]
gberaudo: @b4l I don't see significant time saving with this approach: 10m 40s instead of 11min.

Makefile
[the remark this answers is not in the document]
gberaudo: What did you change to fix it?

api/Dockerfile
b4l: I added the offline flag here as it only applies to a docker build anyways. What do you mean by `local dev`? You can test this by removing & purging everything from previous composition runs and the just use `make run`.

.github/workflows/rust.yml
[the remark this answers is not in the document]
b4l: The CI for Rust takes a little more than 1.5 minutes, the total time is mostly because the UI integration tests.

api/Dockerfile
gberaudo: I mean dev on the developer machine. I just pulled and tried a `make run` but it doesn't rebuild the image so I can not test. (see my comment above about the issue).

gberaudo: I did nothing special since I wanted to test that "make run" automatically take changes in the project and brings up an up-to-date composition.

b4l: This does not work with the changes made here due to docker-compose limitations.

.github/workflows/rust.yml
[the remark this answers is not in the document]
gberaudo: That is good to know, but since we need to wait for the UI to finish anyway, is it worth the drawbacks?

PR #829, opened 2022-02-18
2022.2 translations

anjaamrein:


anjaamrein:
@eleu Could you please control if the pull request worked?

PR #831, opened 2022-02-21
Build improvements

gberaudo:


gberaudo:
Replaces: #827.

b4l (changes requested):
The CI for Rust is not running the test which makes it quite useless. Also this is merely copied from #827 . The latter is also much more refined and runs the test. I don't see why we need this at all.

b4l:
> @b4l, now it runs tests.

It does not run the tests in the CI. Also this requires now a running database to build the Docker image.

.github/workflows/ci.yml
```diff
@@ -21,26 +21,11 @@ jobs:
...
-          docker login -u $DOCKERHUB_USERNAME -p $DOCKERHUB_PASSWORD
-      - name: Build api
-        run: make build_local_api
-
   validate_api:
     runs-on: ubuntu-20.04
     timeout-minutes: 15
     steps:
-      - run: docker system prune --all --force --volumes
+      # - run: docker system prune --all --force --volumes
```
gberaudo, line 28: Do we really want to cache base image?

api/Dockerfile
```diff
@@ -23,21 +24,29 @@ RUN adduser \
...
 
+# Then we copy every files, and detect linting / formating errors
+# This is only detection, fixing the errors should be done outside docker
+# We handle this section in offline mode to avoid unecessary and terribly long crates index update
 COPY ./ .
 ENV SQLX_OFFLINE=true
-RUN cargo clippy --target x86_64-unknown-linux-musl --tests --examples --offline -- -D warnings
 RUN cargo fmt --all -- --check
+RUN cargo clippy --target x86_64-unknown-linux-musl --tests --examples --offline -- -D warnings --no-deps
+RUN cargo test tests --target x86_64-unknown-linux-musl
```
b4l, line 41: This does not run the integration tests in `/tests`

api/Dockerfile
```diff
@@ -23,21 +24,29 @@ RUN adduser \
 WORKDIR /app
 
 # First we handle fetching and building our dependencies
-# We need a stub entry point for that
+# This rarely changes so is done once and for all
+# Cargo needs a stub entry point otherwise it fails
 COPY Cargo.toml Cargo.lock ./
 COPY ./src/stub.rs ./src/main.rs
```
b4l, line 30: There is no need for `stub.rs`, `main.rs` works as well as long as you do not build later before copying the other files. Anyways we build twice now which is quite overkill.

api/DockerfileDev
```diff
@@ -1,10 +1,13 @@
...
+
 FROM rust:latest
 
 RUN rustup target add x86_64-unknown-linux-musl
-RUN apt update && apt install -y musl-tools musl-dev postgresql-client-13
-RUN update-ca-certificates
-RUN cargo install cargo-watch
-RUN cargo install sqlx-cli --no-default-features --features postgres,rustls
+RUN apt update && apt install -y musl-tools musl-dev
+RUN curl -L https://github.com/watchexec/cargo-watch/releases/download/v8.1.2/cargo-watch-v8.1.2-x86_64-unknown-linux-gnu.tar.xz -o cargo-watch.tar.xz
```
b4l, line 8: Don't forget to keep track of updates

docker-compose.yaml
```diff
@@ -1,27 +1,26 @@
...
-      ENVIRONMENT_NAME: local-dev
       PGUSER: &db_user www-data
       PGPASSWORD: &db_pass www-data
       PGDATABASE: &db_name swissgeol-local
       PGHOST: db
       PGPORT: "5432"
       APP_PORT: "3000"
-      SQLX_OFFLINE: "true"
+      DATABASE_URL: "postgres://www-data:www-data@db:5432/swissgeol-local"
     command: ["cargo", "watch", "--shell", "cargo run --offline --target x86_64-unknown-linux-musl"]
```
b4l, line 19: This needs the `--poll` flag

api/DockerfileDev
```diff
@@ -1,10 +1,13 @@
...
+RUN tar xvf cargo-watch.tar.xz
+RUN cp cargo-watch-v8.1.2-x86_64-unknown-linux-gnu/cargo-watch /usr/bin
 
 WORKDIR /app
```
b4l, line 13: You can use `main.rs` and later only `fetch` instead of `build` to get rid of the stub

docker-compose.yaml
b4l: Also running this in offline mode requires to restart the composition on any dependency change.

Makefile
```diff
@@ -8,18 +8,8 @@ run: build_local_api
```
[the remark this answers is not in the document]
b4l: As discussed we only want to run the tests in the composition while using local rust/cargo for the rest.

b4l:
@gberaudo what about having a docker composition that does not run any cargo command on up but a specific command is execute in a later step? So `make run` would be `docker compose up` and then send `cargo watch` to the api container.

api/Dockerfile
gberaudo: We can not use the original `main.rs` because it would cause a rebuild of the dependencies.

I would prefer if cargo was able to build dependencies without the `stub.rs`, but it was not the case when I tested.

The project is built only once:
- the dependencies are built here, which takes ~2min;
- later our api code is built, which takes ~5s.

api/DockerfileDev
gberaudo: Sure, when we want to change to another version of cargo watch we can simply change the version number here.
Idem for the version of the apline image or rust toolchain.

api/DockerfileDev
gberaudo: If we were to use `main.rs` it would rebuild the dependencies very often.

api/Dockerfile
b4l: You can do `RUN echo "fn main() { }" > src/main.rs `

gberaudo:
> @gberaudo what about having a docker composition that does not run any cargo command on up but a specific command is execute in a later step? So `make run` would be `docker compose up` and then send `cargo watch` to the api container.
> 
> We could pass the command as `env` variable in docker compose to additionally have a generic solution.

Do you think of a particular use case?

api/DockerfileDev
b4l: Better use `RUN echo "fn main() {}" > src/main.rs`

b4l:
> Do you think of a particular use case?

Resetting the database came to mind which is not possible while using cargo watch, also this would enable us to run the test with the same composition. Also would potentially make it possible to add an option to use the local api with the composition.

api/Dockerfile
gberaudo: Indeed, it is now fixed, thanks.

gberaudo:
It would be convenient to reset the database each time we start `make run`. That would guarantee a clean state.
I will add a commit for that.

If you want to re-run the tests / run the tests manually / run any oneshot command you can do:
docker-compose exec api "your command and args"

To attach to the api container you can do `docker-compose exec api bash`

> Also would potentially make it possible to add an option to use the local api with the composition.
What do you mean?

b4l (approved):

PR #832, opened 2022-02-22
WIP Add cognito helper

fredj:


b4l:
@fredj you probably want to separate into two structs `CognitoConfig` & `CognitoClient` and then implement `fn new(config: CognitoConfig) -> Self` on `CognitoClient` to do `let client = CgnitoClient::new(config)` then add the Client as Extension layer like PgPool.

fredj:
See #835

PR #833, opened 2022-02-22
Duplicate a topic to a project - Backend

vladyslav-tk:
TODO: Need to solve how we will add AWS creds.

### DIFF
--- a/api/src/handlers.rs
+++ b/api/src/handlers.rs
@@ -1,14 +1,12 @@
-use std::collections::HashMap;
-
+use aws_sdk_s3::Client;
 use aws_sdk_s3::error::CopyObjectError;
 use aws_sdk_s3::output::CopyObjectOutput;
 use aws_sdk_s3::types::SdkError;
-use aws_sdk_s3::Client;
-use axum::extract::Path;
 use axum::{
-    extract::{Extension, Json},
-    http::StatusCode,
+  extract::{Extension, Json},
+  http::StatusCode,
 };
+use axum::extract::Path;
 use serde::{Deserialize, Serialize};
 use sqlx::{FromRow, PgPool};
 use uuid::Uuid;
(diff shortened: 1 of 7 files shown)


api/src/handlers.rs
```diff
@@ -15,3 +64,142 @@ pub async fn health_check(Extension(pool): Extension<PgPool>) -> (StatusCode, St
...
+    {
+        Ok(result) => Ok(axum::Json(result.iter().map(|v: &sqlx::types::Json<Project>| v.to_owned().0).collect())),
+        Err(e) => Err(StatusCode::INTERNAL_SERVER_ERROR),
+    }
+}
+
+#[axum_macros::debug_handler]
+pub async fn get_project(
+    Extension(pool): Extension<PgPool>,
+    Path(params): Path<HashMap<String, String>>,
```
b4l, line 160: this should work `Path(id): Path<Uuid>`, didn't test it though

api/src/handlers.rs
```diff
@@ -15,3 +64,142 @@ pub async fn health_check(Extension(pool): Extension<PgPool>) -> (StatusCode, St
...
+) -> Result<axum::Json<Project>, StatusCode> {
+    let result: Result<sqlx::types::Json<Project>, (StatusCode, String)> = sqlx::query_scalar(
+        r#"SELECT project as "project: sqlx::types::Json<Project>" FROM projects WHERE id = $1"#
+    )
+        .bind::<Uuid>(Uuid::parse_str(params.get("id").ok_or(StatusCode::INTERNAL_SERVER_ERROR).unwrap()).unwrap())
+        .fetch_one(&pool)
+        .await
+        .map_err(internal_error);
```
b4l, line 168: no need too map the error twice here, just use match directly

b4l (approved):
Sorry for the many comments! Nothing big, just skim through. No need too fix everything. I think it is better to merge and iterate.

PR #835, opened 2022-02-25
GSNGM-826: Implement JWT + cognito helper function in the api

b4l:
TODO: Actually use the `Claims` in the protected handlers.


b4l:
This is now quite in line with the specifications of aws: https://docs.aws.amazon.com/cognito/latest/developerguide/amazon-cognito-user-pools-using-tokens-verifying-a-jwt.html

.editorconfig
```diff
@@ -9,3 +9,6 @@ insert_final_newline = true
 
 [*.md]
 trim_trailing_whitespace = false
+
+[.rs]
```
b4l, line 13: this might need a `*`. generally i like to avoid setting up & checking in editor related stuff as it fails to live up to the promise. maybe we can consider removing it altogether?

api/src/auth.rs
```diff
@@ -0,0 +1,121 @@
...
+    /// The cognito client id
+    #[clap(long, env)]
+    pub clientid: String,
+    /// The identity pool id
+    #[clap(long, env)]
+    pub poolid: String,
+}
+
+impl Auth {
+    pub async fn inizialize(&self) -> anyhow::Result<()> {
```
fredj, line 38:
```suggestion
    pub async fn initialize(&self) -> anyhow::Result<()> {
```

.editorconfig
fredj: I've removed the rs part and changed to `*` to `*.js`

PR #838, opened 2022-02-28
Format and rename coordinates in profile

fredj:


eleu:
thanks, this looks good when testing on the demo!

gberaudo (approved):

PR #839, opened 2022-02-28
Remove topic features when going back to list

fredj:


gberaudo (approved):

gberaudo:
@fredj, #841 should fix the npm issue.

PR #848, opened 2022-03-11
upload picture dashboard GeoMol

anjaamrein:
@eleu


eleu:
@fredj this PR probably also needs re-basing on the branch 2022.4_dashboards, could you do that, please? thanks a lot

fredj:
see #850 for then master branch

PR #861, opened 2022-03-25
Misc

b4l:
Selects the newly created view and some project update refactoring. We probably want to create a project store in the future to sync the projects.

### DIFF
--- a/ui/src/elements/ngm-dashboard.ts
+++ b/ui/src/elements/ngm-dashboard.ts
@@ -82,7 +82,6 @@ export class NgmDashboard extends LitElementI18n {
   @state() selectedTopicOrProject: Topic | Project | undefined;
   @state() topics: Topic[] | undefined;
   @state() selectedViewIndx: number | undefined;
-  @state() refreshProjects = true;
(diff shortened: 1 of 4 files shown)


ui/src/elements/ngm-dashboard.ts
```diff
@@ -333,10 +336,10 @@ export class NgmDashboard extends LitElementI18n {
 
     apiClient.duplicateProject(createProject).then(r => r.json())
       .then(id => {
-        apiClient.getProject(id).then(r => r.json()).then(project => {
-          this.selectedTopicOrProject = project;
+        apiClient.getProject(id).then(r => r.json()).then(async project => {
```
gberaudo, line 339: For consistency I suggest to use promise here.

gberaudo (approved):

ui/src/elements/ngm-dashboard.ts
b4l: @gberaudo not sure what you mean by using promise. I made it consistent with line 112.

PR #870, opened 2022-04-07
Add some UI guidelines

gberaudo:


ui/UI_GUIDELINES.md
```diff
@@ -0,0 +1,26 @@
+# UI principles and guidelines
```
b4l, line 1: The content is more about architectural principles and technologies/languages/libraries used. The title is quite misleading, maybe we can find a better one.

ui/UI_GUIDELINES.md
```diff
@@ -0,0 +1,26 @@
...
+
+
+# Pure Typescript
+
+We complement EcmaScript with types defined with Typescript. It allows code completion and provides some level of checking at compilation time.
```
b4l, line 6: So we do a pure mix?

PR #996, opened 2023-03-22
Remove Hotjar code

fredj:


eleu:
thanks, yes we stop using hotjar.

PR #1045, opened 2023-06-07
Query list on click

vladyslav-tk:


ui/src/elements/ngm-object-information.ts
```diff
@@ -44,13 +45,22 @@ export class NgmObjectInformation extends LitElementI18n {
...
+                <div class="ngm-zoom-plus-icon"></div>
+              </button>
+            </div>
+            ${unsafeHTML(it.content)}
+          </div>`) :
         html`
           <table class="ui compact small very basic table">
             <tbody>
-            ${this.info.properties.map(row => {
+            ${this.info.properties?.map(row => {
```
gberaudo, line 63: If `this.info.properties` is `undefined`, it will display "undefined" in the rendered HTML, right?

gberaudo (approved):

PR #1058, opened 2023-06-26
2023.10 11 translations 02

nOester:


nOester:
@eleu 
Could you have a look at the error of this PR. I did some translations, which would be good to be deployed on next thursday 29th of June.
Thanks
Nils

PR #1059, opened 2023-06-26
remove limit=1 in query to api.geo.admin.ch

eleu:


eleu:
cool, thanks I tested and this config looks quite good!

PR #1062, opened 2023-06-30
Allow edit with LV95 coords for points

pedroslvieira:


eleu:
OK thanks, works well in the review link!

PR #1115, opened 2023-10-09
Implement 3D tiles exaggeration

pedroslvieira:


pedroslvieira:
Use 'terrainExaggeration=X' as param for testing

PR #1133, opened 2023-11-06
Add tiles exaggerate with cesium branch updates

pedroslvieira:


pedroslvieira:
use 'terrainExaggeration=X' as param to test

PR #1136, opened 2023-11-09
Update dependencies on the backend

vladyslav-tk:


api/DockerfileDev
```diff
@@ -1,10 +1,10 @@
 # local dev and dev/int/prod images are separate because they
 # are built using a different and incompatible mode (default vs release)
 
-FROM rust:1.70
+FROM rust:1.73
 
 RUN rustup target add x86_64-unknown-linux-musl
-RUN apt update && apt install -y musl-tools musl-dev postgresql-client-13
+RUN apt update && apt install -y musl-tools musl-dev postgresql-client
```
gberaudo, line 7: We should keep the exact version. The rule is that it should be >= the version of the DB we have in production.

api/DockerfileDevMac
```diff
@@ -0,0 +1,19 @@
+## Use a base image that supports Rust and cross-compilation for aarch64
```
gberaudo, line 1: Can you add a comment explaining why this base image is needed on Darwin for local dev but not for building the int/prod images?

gberaudo (approved):

PR #1145, opened 2023-11-20
[Don't merge] Add terrain exaggeration

pedroslvieira:


pedroslvieira:
use 'zExaggeration=X' to test terrain exaggeration, with values between 1 and 100

PR #1152, opened 2023-11-28
[Don't merge] Voxel exaggerate - old rendering

pedroslvieira:


pedroslvieira:
use url param 'zExaggeration=X' to test terrain exaggeration, with values between 1 and 100

PR #1161, opened 2023-12-06
Decrease the number of requests during project geometries edit

vladyslav-tk:


fredj (approved):
Nice use of the decorators !

PR #1240, opened 2024-03-19
GUI for setting URL parameters

vladyslav-tk:


pedroslvieira (commented):
It's cool! Maybe for the layout, the cesium ion button could take the entire width and  have some  y-padding ?

Also, could we add the eye for the zExaggeration as well? It looks like something is missing in the layout, and it could also be useful to, for example, set it to 5x and be able to toggle it visible/invisible

image

pedroslvieira (approved):

pedroslvieira:
Oh ok, I did not see that there is a mock-up
Yeah that's weird, I'm using chrome and even on dev now I see no padding.

PR #1243, opened 2024-03-22
Exaggeration improvements

vladyslav-tk:


ui/src/cesium-toolbar.ts
```diff
@@ -227,6 +280,50 @@ export class CesiumToolbar extends LitElement {
                    this.viewer!.resolutionScale = Number(evt.target.value);
                    this.currentScale = this.viewer!.resolutionScale;
                }}>
+      </div>
+      <div class="divider"></div>
+      <div>
+        Enable shadows
```
gberaudo, line 286: Shouldn't all these be translatable?

PR #1267, opened 2024-05-02
2024.07 translation

nOester:


fredj:
todo @fredj: also cherry pick this to 2024.07.0 tag

PR #1276, opened 2024-05-24
Update layertree.ts - Make public boreholes private

nOester:
The boreholes shall be accessible only for users in the "ngm-prod-privileged"-group. After re-integration of boreholes it will accessable publicly.

cf. GSNGM-1156


nOester:
Boreholes are switched off, but Login does not work

vladyslav-tk (approved):

PR #1282, opened 2024-05-27
Fix 'make run'

vladyslav-tk:


tyrossel:
While running `make run`, I run into another error due to to `docker-compose`, which is deprecated.

I would replace `docker-compose` by the new `docker compose`

vladyslav-tk:
> While running `make run`, I run into another error due to to `docker-compose`, which is deprecated.
> 
> I would replace `docker-compose` by the new `docker compose`

yes, thanks, I updated it

tyrossel (approved):

PR #1391, opened 2024-11-20
Feature 1365: Behebe Warnungen im Backend

### LINKED ISSUE #1365: Behebe Warnungen im Backend
**Beschreibung**

Im backend gibt es Warnungen. Wir würden gerne die Code Quality Regeln auf `Treat warnings as Errors` stellen.
Behebe dafür die bestehenden Warnungen.

**Acceptance Criteria**
- Es gibt keine Warnungen im Backend

daniel-va:
Resolves #1365.


api/src/handlers.rs
```diff
@@ -85,26 +85,28 @@ pub struct Member {
 
 #[derive(Serialize, Deserialize, Clone, Debug, FromRow)]
 #[allow(non_snake_case)]
```
TIL-EBP, line 87: do we still need this?

api/src/handlers.rs
```diff
@@ -596,5 +598,5 @@ fn generate_asset_name() -> String {
         .take(40)
         .map(char::from)
         .collect();
-    return format!("{}_{}.kml", Utc::now().timestamp(), rand_string);
+    format!("{}_{}.kml", Utc::now().timestamp(), rand_string)
```
TIL-EBP, line 601: is this the prefered way to do it in rust?

TIL-EBP (approved):

api/src/handlers.rs
daniel-va: Yes, in Rust, Functions always return their last expression by default. `return` should only be used for early returns.

api/src/handlers.rs
daniel-va: No, I forgot to check if there were any other suppressed warnings.

PR #1393, opened 2024-11-25
Fix: sonar cloud quick fixes

### LINKED ISSUE #1340: SonarCloud Issues
**Beschreibung**
Die in SonarCloud bestehenden Issues sollen bearbeitet werden:
[Summary - swissgeol-viewer-app in swisstopo SonarCloud](https://sonarcloud.io/summary/overall?id=swisstopo_swissgeol-viewer-app)

**ToDos**
- [ ] Prüfen welche SonarCloud-Issues relevant sind
- [ ] Wenn geprüft dann mit @nOester  besprechen und relevante issues identifizieren
Dann Folge tasks:
- [ ] Relevante Issues bearbeiten (Quickwins und nicht relevante abhaken)

**Acceptance criteria**
- Es ist definiert und begründet(in diesem Issue) welche SonarCloud welche Issues relevant sind und welche nicht
- Die Relevanten Issues sind behoben oder zur Umsetzung geplant

TIL-EBP:
resolves #1340 quick wins

### DIFF
--- a/ui/src/elements/ngm-object-information.ts
+++ b/ui/src/elements/ngm-object-information.ts
@@ -22,13 +22,8 @@ export class NgmObjectInformation extends LitElementI18n {
     this.hidden = !this.opened;
 
     QueryStore.objectInfo.subscribe(info => {
-      if (info) {
-        this.info = info;
-        this.opened = !!info;
-      } else if (this.opened) {
-        this.opened = false;
-        this.info = undefined;
-      }
+      this.info = info;
+      this.opened = !!info;
     });
 
     document.addEventListener('keydown', (event) => {
--- a/api/src/database.rs
+++ b/api/src/database.rs
@@ -1,4 +1,3 @@
-use serde::Serialize;
 use sqlx::{
     postgres::{PgConnectOptions, PgPoolOptions},
     Connection, Executor, PgConnection, PgPool,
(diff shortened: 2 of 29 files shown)


ui/src/elements/ngm-object-information.ts
```diff
@@ -1,119 +1,119 @@
...
+    this.opened = false;
+    this.hidden = !this.opened;
+
+    QueryStore.objectInfo.subscribe(info => {
+      if (info) {
+        this.info = info;
+        this.opened = true;
+      } else if (this.opened) {
+        this.opened = false;
+        this.info = undefined;
```
daniel-va, line 30: This could be simplified much further.
```suggestion
      this.info = info;
      this.opened = info == null;
```

daniel-va (changes requested):

ui/src/elements/ngm-object-information.ts
TIL-EBP: I had your refactored version before, but since this is an` else if `and not just an `else`, it is not quite equivalent.
It may never occur, but in the new version` this.info` is set to `undefined `if `info `is `undefined `and` this.opened `is `false`. In the original version, `this.info` will only be set to `undefined`, if `this.opened` is `true`.
This case may never occur, but I am not sure of it, which is why i did not simplify all the way

daniel-va: As far as I can see, the `else if` is a very premature optimization - it is there so that `info` and `opened` do no get set to the same values that they already contain.

daniel-va (approved):

PR #1407, opened 2025-01-07
Feature 1367: Datenkatalog Tabs erstellen

### LINKED ISSUE #1367: Datenkatalog Tap erstellen
**Beschreibung**
Der untere Bereich des Datenpanels soll in drei Taps aufgeteilt werden:
- Datakatalog
- Upload
- Einstellungen

Der Datenkatalog Tap soll den Bestehenden Datenkkatalog enthalten

**ToDos**
- [ ] Der Datenkatalog ist gemäss https://www.figma.com/design/cEiOoOazAQZqpRY92ZhBeO/swissgeol?node-id=9583-8200&node-type=canvas&t=PhhUf89SndRhJUXl-0 in ein Tap integriert

**Acceptance Criteria**
- [ ] tbd

daniel-va:
Resolves #1367.


ui/src/layers/ngm-layers-item.ts
vej-ananas, line 1: The overflow and scrolling behavior is not correct yet.
When we add many layers (like 20), the expected layout should still be half the page for each, the layers and the catalogue rather than having a big scrollbar. Each of the two should have their own scrollbar

daniel-va: I adjusted it now that the two parts scale independently, and have individual scroll bars. In my opinion, there are some readability issues when both scroll bars are in use - might be necessary to check with the UX team on how to proceed with that.

vej-ananas: I think both sections need `max-height: 50%;` and then it's perfect

PR #1413, opened 2025-01-13
Feature 1368: Upload-Tab anpassen

### LINKED ISSUE #1368: Upload Tab anpassen
**Beschreibung**
Das Design der Inhalte des  Tab "Upload" soll ensprechend Figma angepasst werden.
https://www.figma.com/design/cEiOoOazAQZqpRY92ZhBeO/swissgeol?node-id=9583-8200&node-type=canvas&t=PhhUf89SndRhJUXl-0

**Acceptance Criteria**
- [ ] Die Funktionalität bleibt erhalten, das Design ist angepasst.

daniel-va:
Resolves #1368.
Depends on #1412.


ui/src/components/core/core-checkbox.ts
```diff
@@ -19,10 +29,17 @@ export class CoreCheckbox extends LitElement {
   readonly render = () => html`
     <label @click="${this.handleClick}">
       <input type="checkbox" ?checked="${this.isActive}">
-      <div class="icon"></div>
-      ${this.label == null
-        ? ''
-        : html`<span class="label">${this.label}</span>`}
+      <div class="icon">
+        <svg width="12" height="10" viewBox="0 0 12 10" fill="none" xmlns="http://www.w3.org/2000/svg">
```
TIL-EBP, line 33: why don't you use a core-icon here?

ui/src/components/core/core-modal.ts
```diff
@@ -0,0 +1,117 @@
...
+      width: 326px;
+    }
+
+    dialog::backdrop {
+      background-color: #111827B2;
+      opacity: 0.7;
+    }
+
+    dialog > div {
+      padding: 24px;
```
TIL-EBP, line 110: I removed this here and added it in the uplaod modal itself, because in the disclaimer, i need the line breaks to span the entire width of the modal

ui/src/components/layer/upload/layer-upload-kml.ts
```diff
@@ -20,42 +22,92 @@ export default class NgmLayerUploadKml extends LitElementI18n {
...
+
+  firstUpdated() {
+    const handleDragStart = (e: DragEvent) => {
+      if (e.target !== this.uploadButton) {
+        return;
+      }
+      this.uploadButton.classList.add('is-active');
+    };
+    const handleDragEnd = (e: DragEvent) => {
+      console.log(e);
```
TIL-EBP, line 47: remove console log. Also, the drag and drop does not seem to work properly. It opens the files dropped instead of opening the modal. `handleDrop` is missing here

ui/src/components/layer/upload/layer-upload-kml.ts
```diff
@@ -20,42 +22,92 @@ export default class NgmLayerUploadKml extends LitElementI18n {
...
-  private accessor uploadKmlInput!: HTMLInputElement;
+    this.handleUpload = this.handleUpload.bind(this);
+  }
+
+  firstUpdated() {
+    const handleDragStart = (e: DragEvent) => {
+      if (e.target !== this.uploadButton) {
+        return;
+      }
+      this.uploadButton.classList.add('is-active');
```
TIL-EBP, line 44: This does not always work perfectly for some reason (is-active seems to be removed when i drag the file far into the drop zone

ui/src/components/layer/upload/layer-upload-kml.ts
```diff
@@ -20,42 +22,92 @@ export default class NgmLayerUploadKml extends LitElementI18n {
...
+    this.modal = CoreModal.open({size: 'small'}, html`
+      <ngm-layer-upload-kml-modal
+        .file="${file}"
+        @confirm="${this.handleUpload}"
+        @cancel="${() => this.modal?.close()}"
+      ></ngm-layer-upload-kml-modal>
+    `);
+  }
+
+  private validateFile(file: File | null): boolean {
```
TIL-EBP, line 87: I never reach this when dropping a file of a wrong type (related to not handling the file upload on drag and drop)

TIL-EBP (changes requested):
Appart from the drag and drop, it looks fine

ui/src/components/core/core-modal.ts
daniel-va: Maybe we should consider making this configurable on the modal, e.g. give it a `hasNoPadding` property? If we remove it anywhere, it might become a hassle with us having to add the same padding to every modal body that needs it.

TIL-EBP: Yeah that is a good idea, i can add it in my PR

ui/src/components/layer/upload/layer-upload-kml-modal.ts
```diff
@@ -0,0 +1,92 @@
...
+        file: this.file,
+        isClampEnabled: this.isClampEnabled,
+      }
+    }));
+  }
+
+  render = () => html`
+    <h2>
+      ${i18next.t('dtd_upload_modal_title')}
+    </h2>
```
TIL-EBP, line 35: why not use an h1?

TIL-EBP (changes requested):

ui/src/components/layer/upload/layer-upload-kml-modal.ts
daniel-va: Because `h1` is the top level element and [should only be used once per page](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Heading_Elements#avoid_using_multiple_h1_elements_on_one_page).
`h2` is a good middle ground for a section that stands on its own, but is not the page's main heading.

ui/src/components/core/core-checkbox.ts
daniel-va: Because the icon used in the checkbox is not part of [icons list](https://www.figma.com/design/cEiOoOazAQZqpRY92ZhBeO/swissgeol?node-id=32-1169&p=f&t=FS27sP1vwe8hv6tv-0), and will only be used here. This way, we do not create any confusion over which checkmark to use, and don't need to deal with re-styling the nested `ngm-core-icon` within the checkbox.

ui/src/components/layer/upload/layer-upload-kml.ts
```diff
@@ -85,16 +100,17 @@ export default class NgmLayerUploadKml extends LitElementI18n {
   }
 
   private validateFile(file: File | null): boolean {
+    console.log(file);
```
TIL-EBP, line 103: remove

ui/src/components/layer/upload/layer-upload-kml.ts
```diff
@@ -85,16 +100,17 @@ export default class NgmLayerUploadKml extends LitElementI18n {
...
       return false;
     }
     if (!file.name.toLowerCase().endsWith('.kml')) {
-      showBannerError(this.toastPlaceholder, i18next.t('dtd_file_not_kml'));
+      this.violation = i18next.t('dtd_file_not_kml');
       return false;
     }
     if (this.isFileTooLarge(file)) {
-      showBannerError(this.toastPlaceholder, `${i18next.t('dtd_max_size_exceeded_warn')} ${this.maxFileSize}MB`);
+      this.violation = `${i18next.t('dtd_max_size_exceeded_warn')} ${this.maxFileSize}MB`;
```
TIL-EBP, line 113: remove the violations after a successful upload of another file

ui/src/components/layer/upload/layer-upload-kml.ts
```diff
@@ -85,16 +100,17 @@ export default class NgmLayerUploadKml extends LitElementI18n {
...
       return false;
     }
     return true;
```
TIL-EBP, line 116:
```suggestion
    this.violation = null
    return true;
```

TIL-EBP (approved):
Works fine now! I just found two minor things

PR #1417, opened 2025-01-14
Feature 1375: disclaimer

### LINKED ISSUE #1375: Disclaimer umbauen
**Beschreibung**
Der bestehende Disclaimer soll entsprechend der Definition in Figma: https://www.figma.com/design/cEiOoOazAQZqpRY92ZhBeO/swissgeol?node-id=9598-111197&node-type=section&t=EyWoIe28GOV1xAbK-0 umgebaut werden.

**ToDos**
- [ ] Umbau bestehender Disclaimer in die in Figma (s. oben) definierte Form. Der Text des Disclaimers ist im translation-key `"tracking_text"` der jeweilgen Übersetzungs-files in https://github.com/swisstopo/swissgeol-viewer-app/tree/develop/ui/locales

**Acceptance Criteria**
- [ ] Disclaimer entspricht der Definition in Figma (s. oben)

TIL-EBP:
resolves #1375


ui/src/components/core/core-modal.ts
```diff
@@ -51,6 +53,9 @@ export class CoreModal extends LitElement {
   @property({type: Boolean})
   accessor isPersistent = true;
 
+  @property(({type: Boolean, reflect: true}))
```
daniel-va, line 56: So far, I always changed the reflected attribute to more closely resemble a normal HTML attribute. We haven't discussed this yet, so maybe we should talk about this sometime?
```suggestion
  @property(({type: Boolean, attribute: 'no-padding', reflect: true}))
```

ui/src/components/core/core-modal.ts
```diff
@@ -109,9 +118,14 @@ export class CoreModal extends LitElement {
     dialog > div {
       padding: 24px;
     }
+
+    :host([hasNoPadding]) > dialog > div {
+      padding: 0;
+    }
```
daniel-va, line 124: This could be shortened - not essential, but is IMO easier to read and reason about this way.
```suggestion
    :host(:not([hasNoPadding])) dialog > div {
      padding: 24px;
    }
```

ui/src/components/language-selector/ngm-language-selector.ts
```diff
@@ -0,0 +1,100 @@
...
+  firstUpdated(): void {
+    if (this.shadowRoot != null) {
+      $(this.shadowRoot.querySelectorAll('.ui.dropdown')).dropdown({
+        on: 'mouseup',
+        collapseOnActionable: false,
+      });
+    }
+  }
+
+  static readonly styles = css`
```
daniel-va, line 25: Switch order of `styles` and `render`.

ui/src/components/language-selector/ngm-language-selector.ts
```diff
@@ -0,0 +1,100 @@
...
+    }
+
+    .ngm-lang-dropdown:hover {
+      background-color: var(--color-hovered);
+      color: var(--color-main--dark);
+    }
+
+    .ngm-lang-dropdown.active ngm-icon[icon="dropdown"] {
+      transform: rotate(180deg);
+    }
```
daniel-va, line 46: This rule currently doesn't do anything, as the dropdown icon is now based on `ngm-core-icon`.

ui/src/components/language-selector/ngm-language-selector.ts
```diff
@@ -0,0 +1,100 @@
...
+      $(this.shadowRoot.querySelectorAll('.ui.dropdown')).dropdown({
+        on: 'mouseup',
+        collapseOnActionable: false,
+      });
+    }
+  }
+
+  static readonly styles = css`
+
+    .ngm-lang-dropdown {
```
daniel-va, line 27: IMO you could just drop the `ngm-lang-` prefix for the classes here. Since this is now in the Shadow DOM, they will not conflict with anything anyways.

ui/src/components/language-selector/ngm-language-selector.ts
```diff
@@ -0,0 +1,100 @@
...
+
+    .ngm-lang-item {
+      display: flex;
+      justify-content: space-between;
+      align-items: center;
+      padding: 8px 12px;
+      font-size: 14px;
+      font-family: var(--font);
+      line-height: 20px;
+      letter-spacing: calc(14px * 0.0025);
```
daniel-va, line 63: Use `applyTypograpy` instead of specifying text properties explicitly.

ui/src/ngm-app.ts
```diff
@@ -125,13 +121,16 @@ export class NgmApp extends LitElementI18n {
...
   private resolutionScaleRemoveCallback: Event.RemoveCallback | undefined;
+  private readonly disclaimer: CoreModal | null = null;
 
   @consume({context: clientConfigContext})
   accessor clientConfig!: ClientConfig;
 
   constructor() {
     super();
 
+    this.disclaimer = CoreModal.open({isPersistent: true, size: 'large', hasNoPadding: true}, html`<ngm-tracking-consent @confirm="${(event) => this.onTrackingAllowedChanged(event)}"></ngm-tracking-consent>`);
```
daniel-va, line 132: Split this across multiple lines, + pass `onTrackingAllowedChanged` directly.
```suggestion
    this.disclaimer = CoreModal.open({isPersistent: true, size: 'large', hasNoPadding: true}, html`
      <ngm-tracking-consent
        @confirm="${this.onTrackingAllowedChanged}"
      ></ngm-tracking-consent>
    `);
```

ui/src/ngm-app.ts
```diff
@@ -125,13 +121,16 @@ export class NgmApp extends LitElementI18n {
   private queryManager: QueryManager | undefined;
   private waitForViewLoading = false;
   private resolutionScaleRemoveCallback: Event.RemoveCallback | undefined;
+  private readonly disclaimer: CoreModal | null = null;
```
daniel-va, line 124: I probably would prefer if this wasn't `readonly`, and instead was set to `null` once it is closed. This way, we keep the object reference throughout the application's entire lifetime.

ui/src/components/tracking-consent/tracking-consent-modal.ts
```diff
@@ -0,0 +1,92 @@
...
+          </div>
+          <div class="footer">
+            <ngm-core-button @click="${() => this.accept()}">
+              ${i18next.t('accept')}
+            </ngm-core-button>
+          </div>
+      `;
+  }
+
+  private accept() {
```
daniel-va, line 85: Move this above `render`.

ui/src/components/tracking-consent/tracking-consent-modal.ts
```diff
@@ -0,0 +1,92 @@
+import {css, html} from 'lit';
```
daniel-va, line 1: Adjust this file's name so it matches the class within it.

ui/src/components/language-selector/ngm-language-selector.ts
```diff
@@ -0,0 +1,100 @@
+import {css, html, unsafeCSS} from 'lit';
```
daniel-va, line 1: I don't really like creating a lot of directories within `src/components/` that only contain a single thing - Maybe we can move `ngm-language-selector` and `ngm-tracking-consent` into the same directory? Something like `navigation` or `layout`?

ui/src/components/language-selector/ngm-language-selector.ts
```diff
@@ -0,0 +1,100 @@
...
+      letter-spacing: calc(14px * 0.0025);
+    }
+
+    .ngm-lang-item:hover {
+      background-color: var(--color-pressed);
+    }
+
+    .ngm-lang-title .ngm-dropdown-icon {
+      background-color: var(--ngm-interaction);
+      transform: rotate(90deg);
```
daniel-va, line 72: The color variables used here are not part of the standardized Figma colors. Either use an equivalent color variable from that set, or move the variable to them, in case it is missing.

ui/src/components/language-selector/ngm-language-selector.ts
```diff
@@ -0,0 +1,100 @@
...
+    }
+    ${unsafeCSS(fomanticTransitionCss)}
+    ${unsafeCSS(fomanticDropdownCss)}
```
daniel-va, line 75: These should probably be at the top of the CSS, so they don't overwrite any of our own rules by accident.

ui/src/components/language-selector/ngm-language-selector.ts
```diff
@@ -0,0 +1,100 @@
...
+          <div class="ui dropdown ngm-lang-dropdown">
+          <div class="ngm-lang-title">
+            ${i18next.language?.toUpperCase()}
+            <ngm-core-icon icon="dropdown" />
+          </div>
+          <div class="menu">
+            ${SUPPORTED_LANGUAGES.map(lang => html`
+              <div class="item" @click="${() => i18next.changeLanguage(lang)}" style="padding: 0; width: 85px;">
+                <div class="ngm-lang-item">
+                  <ngm-core-icon style="${styleMap({'visibility': i18next.language?.toUpperCase() === lang?.toUpperCase() ? 'visible' : 'hidden'})}" icon="checkmark"></ngm-core-icon>
```
daniel-va, line 91: Not a big fan of applying styles directly - maybe move these to a class?

ui/src/components/tracking-consent/tracking-consent-modal.ts
```diff
@@ -0,0 +1,92 @@
...
+@customElement('ngm-tracking-consent')
+export class NgmTrackingConsent extends LitElementI18n {
+  @state()
+  accessor isAllowed: boolean = true;
+  @state()
+  accessor isAccepted: boolean = false;
+
+  constructor() {
+    super();
+  }
```
daniel-va, line 19: Remove empty constructor.
```suggestion
```

ui/src/components/tracking-consent/tracking-consent-modal.ts
```diff
@@ -0,0 +1,92 @@
...
+    }
+
+    .footer {
+      display: flex;
+      justify-content: flex-end;
+      padding-top: 24px;
+    }
+  `;
+
+  render() {
```
daniel-va, line 64: Move this above `styles`.

ui/src/components/tracking-consent/tracking-consent-modal.ts
```diff
@@ -0,0 +1,92 @@
...
+  @state()
+  accessor isAccepted: boolean = false;
+
+  constructor() {
+    super();
+  }
+
+  static readonly styles = css`
+    :host  > div {
+      padding: 24px;
```
daniel-va, line 23: Maybe we want to standardize `24px` as something like `--modal-padding`?

ui/src/components/tracking-consent/tracking-consent-modal.ts
```diff
@@ -0,0 +1,92 @@
...
+
+    h2 {
+      ${applyTypography('modal-title')}
+      margin: 0;
+    }
+
+    h3 {
+      ${applyTypography('modal-title')}
+      font-size: 16px;
+      margin: 0 0 16px;
```
daniel-va, line 34: Now that we apparently have two levels of modal titles, we should consider adding a new typography, so we have `modal-title-1` and `modal-title-2` (analog to body and headings).

ui/src/components/tracking-consent/tracking-consent-modal.ts
```diff
@@ -0,0 +1,92 @@
...
+
+    }
+
+    .header {
+      display: flex;
+      justify-content: space-between;
+      align-items: center;
+      padding-bottom: 24px;
+
```
daniel-va, line 43: Unnecessary line break.
```suggestion
```

ui/src/components/core/core-modal.ts
TIL-EBP: When do we use a '-' an when not? For example, 'isactive' is used without

ui/src/ngm-app.ts
TIL-EBP: When i pass `onTrackingAllowedChanged ` directly, i get an nerror when clicking the accept-button, that `showNavigationHint` is not a function

ui/src/components/core/core-modal.ts
daniel-va: I would:
- drop `is` or `has` prefix, if present
- insert `-` before each uppercase character
So `isactive` should probably be `active`.

ui/src/components/tracking-consent/tracking-consent-modal.ts
TIL-EBP: I left the file name as it, but adjusted the class name and tagName instead

ui/src/components/tracking-consent/tracking-consent-modal.ts
TIL-EBP: should we then put all padding values in global variabels? such as padding-s to padding-xl or somethink like that?

ui/src/components/language-selector/ngm-language-selector.ts
TIL-EBP: I agree that we should collect them somewhere, bubt I don't really like either of the suggsetions. What about `general`

ui/src/components/tracking-consent/tracking-consent-modal.ts
daniel-va: Depends on how consistent/clear they are in the Figma - this is the first one that makes sense to extract for me, as we will use it in multiple locations.

Just because a property is the same in two places doesn't mean that these values are related (e.g. the padding of the heading and the padding of the button could both be 8px, but they clearly don't have any relation to each other). The benefit of a variable is that it makes it both easier to change the padding in the future, but it also documents that the padding has this size because it represents a modal.

ui/src/components/language-selector/ngm-language-selector.ts
daniel-va: I mean, there is `core` already, which is used for basic ui components, and names like `shared` would conflict with that (I would be confused what the difference between the two is).

In the future, we will also need a directory for things like the navigation and side bar - this should probably be the same directory that these two components are in. IMO either we store them in `core`, or use a name that clearly separates them from `core`, like `layout`, `scaffold` or `navigation` - or anything else that is not as interchangeable with `core`.

TIL-EBP: Okay, then I would put them all in `layout`. I dont' wnat to put them into `core`, because they are very different IMO.

ui/src/components/layout/language-selector.ts
```diff
@@ -33,68 +57,45 @@ export class NgmLanguageSelector extends LitElementI18n {
...
-      color: var(--color-main);
+      color: var(--color-primary);
     }
 
-    .ngm-lang-dropdown:hover {
-      background-color: var(--color-hovered);
-      color: var(--color-main--dark);
+    .container:hover {
+      background-color: var(--color-secondary--hovered);
+      color: var(--color-medium-emphasis);
```
daniel-va, line 65: This is the old color variable, the new one is `--color-text--emphasis--medium` (could just globally replace this, as it is the exact same value).

ui/src/components/layout/tracking-consent-modal.ts
```diff
@@ -60,33 +82,10 @@ export class NgmTrackingConsent extends LitElementI18n {
...
 
-  private accept() {
-    this.dispatchEvent(new CustomEvent('confirm', {
-      detail: {
-        allowed: this.isAllowed
-      }
-    }));
-  }
+export interface TrackingConsentModalEventDetail {
+  allowed: boolean;
```
daniel-va, line 90: Add `is` prefix for booleans.
```suggestion
  isAllowed: boolean;
```

daniel-va (approved):

PR #1420, opened 2025-01-15
Feature/viewer 1369 einstellungen tab

### LINKED ISSUE #1369: Einstellungen Tab anpassen
**Beschreibung**
Das Design der Inhalte des  Tab "Einstellungen" soll ensprechend Figma angepasst werden.
https://www.figma.com/design/cEiOoOazAQZqpRY92ZhBeO/swissgeol?node-id=9583-8200&node-type=canvas&t=PhhUf89SndRhJUXl-0

**Acceptance Criteria**
- [ ] Die Funktionalität bleibt erhalten, das Design ist angepasst.

TIL-EBP:
resolves #1369


ui/src/components/layer/options/layer-options.ts
```diff
@@ -54,12 +58,12 @@ export class NgmLayerOptions extends LitElementI18n {
     this.viewer?.scene.requestRender();
   }
 
-  private updateExaggeration(evt: InputEvent) {
+  private updateExaggeration(evt: CustomEvent) {
```
daniel-va, line 61: Use a specific event type here.

ui/src/components/layer/options/layer-options.ts
```diff
@@ -68,45 +72,71 @@ export class NgmLayerOptions extends LitElementI18n {
...
-          const exaggeration = this.hideExaggeration ? 1 : this.exaggeration;
-          this.viewer.scene.verticalExaggeration = exaggeration;
-          this.updateExaggerationForKmls();
-          NavToolsStore.exaggerationChanged.next(exaggeration);
-          this.viewer.scene.requestRender();
-        }}>HIDE
+      <div class="group">
+        <ngm-core-icon
+          icon="${this.hideExaggeration ? 'invisible' : 'visible'}"
+          title=${!this.hideExaggeration ? i18next.t('dtd_hide_exaggeration') : i18next.t('dtd_show_exaggeration')}
```
daniel-va, line 78: Invert this ternary.
```suggestion
          title=${this.hideExaggeration ? i18next.t('dtd_show_exaggeration') : i18next.t('dtd_hide_exaggeration')}

```

ui/src/components/layer/options/layer-options.ts
```diff
@@ -68,45 +72,71 @@ export class NgmLayerOptions extends LitElementI18n {
...
-          <label>${i18next.t('dtd_exaggeration_map')}</label>
-          <label>${(this.exaggeration).toFixed()}x</label>
+      <hr>
+      <div class="group">
+        <ngm-core-slider
+          .min="${1}"
+          .max="${20}"
+          .step="${1}"
+          .value="${this.exaggeration}"
+          @change=${(evt: CustomEvent) => this.updateExaggeration(evt)}
```
daniel-va, line 90:
```suggestion
          @change=${this.updateExaggeration}

```

ui/src/components/layer/options/layer-options.ts
```diff
@@ -68,45 +72,71 @@ export class NgmLayerOptions extends LitElementI18n {
...
+          @pointerup="${debounce(() => this.updateExaggerationForKmls(), 300)}"
```
daniel-va, line 91:
```suggestion
          @pointerup="${debounce(this.updateExaggerationForKmls, 300)}"

```

ui/src/components/layer/options/layer-options.ts
```diff
@@ -68,45 +72,71 @@ export class NgmLayerOptions extends LitElementI18n {
...
+        ></ngm-core-slider>
+        <div class="chip-container">
+          <ngm-core-chip >${(this.exaggeration).toFixed()}x</ngm-core-chip>
```
daniel-va, line 94:
```suggestion
          <ngm-core-chip>${this.exaggeration.toFixed()}x</ngm-core-chip>

```

ui/src/components/layer/options/layer-options.ts
```diff
@@ -68,45 +72,71 @@ export class NgmLayerOptions extends LitElementI18n {
...
+    .group {
+      display: flex;
+      justify-content: flex-start;
+      gap: 6px;
+      align-items: center;
+      margin: 10px;
+    }
+
+    ngm-core-slider {
+      flex-grow: 1;
```
daniel-va, line 132: I would prefer `width: 100%` over `flex-grow: 1` - it has the same effect, but makes it clear what it is supposed to achieve.

ui/src/components/layer/options/layer-options.ts
```diff
@@ -68,45 +72,71 @@ export class NgmLayerOptions extends LitElementI18n {
...
+      display: flex;
```
daniel-va, line 133: This should probably be defined directly in the `ngm-core-slider` component, as it centers its contents, which should always be the case when we use a slider.

daniel-va (changes requested):

ui/src/components/core/core-slider.ts
```diff
@@ -0,0 +1,104 @@
...
+  @property({type: Number})
+  accessor max: number = 10;
+
+  @property({type: Number})
+  accessor step: number = 1;
+
+  @property({type: Number})
+  accessor value: number = 0;
+
+  onInputChange(evt: InputEvent) {
```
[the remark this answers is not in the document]
TIL-EBP: I would suggest event, I am not a huge fan of one-letter-variables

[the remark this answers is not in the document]
daniel-va: Fine by me, so we use `event` for every event then 👍

For the naming of event handlers, the reason I use `handle*` is that I like methods to be verbs, so they tell me what they do. So if you don't care either way, I would say we use that.

ui/src/components/core/core-slider.ts
```diff
@@ -0,0 +1,104 @@
+import {css, html, LitElement, unsafeCSS} from 'lit';
+import {customElement, property} from 'lit/decorators.js';
+import './core-icon';
+import sliderCss from '../../style/ngm-slider.css';
```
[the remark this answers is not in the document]
TIL-EBP: I am currently not using any styles from this file. Owever, I also did not implement any of the element states so far because i did not find any for the slider in the Figma. I also don't want to use the old ones because they do not fit with the new design.

daniel-va (approved):

PR #1424, opened 2025-01-27
Feature/viewer 1418 add prettier

### LINKED ISSUE #1418: Add Prettier as linter
**Description**
Add Prettier as Linter and run it for the whole project.
Add it to the CI/CD pipeline as check.

Configure the following:
- Boolean properties start with `is`
- Enforce order inside component:
	Properties
	Methods
	Render
	Css

TIL-EBP:
resolves #1418 

Since so many files have changed, here is a list of the relevant ones: 

- eslint.config.mjs
- .prettierrc
- .prettierignore
- package.json


ui/eslint.config.mjs
```diff
@@ -1,158 +1,208 @@
...
+    // plugins: { 'sort-class-members': sortClassMembers },
+    rules: {
+      ...config.rules,
+      ...sharedConfig.rules,
+      '@typescript-eslint/naming-convention': [
+        'error',
+        {
+          selector: [
+            'variable',
+            // 'autoAccessor', // This currently applies to all accessors, not just boolean ones
```
TIL-EBP, line 137: For some reason, this selector will then apply the check to all the properties of a lit `customElement`, not just boolean ones. 
I have not yyet figured out why this is the case. So currently, the naming convention is not applied to class properties

TIL-EBP: Yeah, I found the same. I still find it weird because in the [Docs](https://typescript-eslint.io/rules/naming-convention/), the define allowed types...
So do we just leave it out and try to remember to name the properties accordingly? OR should we try to find another way?

daniel-va: I'm perfectly fine with just leaving it out - in many projects, conventions like this are not enforced by a linter anyways.

ui/package.json
```diff
@@ -3,7 +3,9 @@
   "private": true,
   "type": "module",
   "scripts": {
-    "lint": "eslint",
+    "lint": "npx prettier --check . && npx eslint",
+    "lint:fix": "npm run lint -- --fix",
+    "prettier": "npx prettier . --write",
```
daniel-va, line 8: I have some issues with these commands:
- `lint` runs both prettier and eslint, but `lint:fix` runs only eslint
- I would like a command that fixes both eslint and prettier - either move `prettier` into `lint:fix` or add a new command that runs both.
- prettier should probably run after eslint, as it is has final say over which formatting is correct.

TIL-EBP: These commands are the same as in assets, so if we change them here, we should also change them there. 
I can add a `format` command, that runs `lint:fix` and `prettier --write` (same as in assets)

daniel-va (approved):

PR #1449, opened 2025-02-17
Fix: Layer Catalog Height

daniel-va:
Fixes the layer catalog not taking up as much space as available, effectively leaving half of the sidebar empty when no additional layers are selected.

### DIFF
--- a/ui/src/components/layer/layer-panel.ts
+++ b/ui/src/components/layer/layer-panel.ts
@@ -1,7 +1,6 @@
-import { LitElementI18n } from 'src/i18n';
 import { css, html } from 'lit';
 import i18next from 'i18next';
-import { customElement, property } from 'lit/decorators.js';
+import { customElement, property, query } from 'lit/decorators.js';
 import { LayerConfig } from 'src/layertree';
 import {
   LayerEvent,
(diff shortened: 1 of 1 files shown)


TIL-EBP (changes requested):
When opening all the accordions in the layer catalogue, the soace for the active layers shrinks until they are no longer visible. 
image

ui/src/components/layer/layer-panel.ts
```diff
@@ -31,23 +39,37 @@ export class LayerPanel extends LitElementI18n {
     super.connectedCallback();
     this.setAttribute('role', 'complementary');
   }
+
+  firstUpdated(): void {
+    const observer = new ResizeObserver(() => {
+      console.log('okay?');
```
TIL-EBP, line 45: remove ;)

TIL-EBP (changes requested):
This works up to a point where I add as many layers as I can, which leads to the catalogue disappearing completely. Also, it does feel a little overcomplicated. 


I think the whole thing could be achieved by the following: 
Remove the `max-height` from the section selector (as you did in this change)
give the tab div a `flex: 1` and the layers div a `max-height: 50%`

TIL-EBP (approved):

PR #1450, opened 2025-02-17
Fix: Layer Grab Indicator

daniel-va:
Fix cursor not showing dragging icon while a layer item is being reordered.

### DIFF
--- a/ui/src/components/core/core-button.ts
+++ b/ui/src/components/core/core-button.ts
@@ -101,7 +101,7 @@ export class CoreButton extends LitElement {
         color: var(--button-text--pressed);
         background-color: var(--button-bg--pressed);
         border-color: var(--button-border--pressed);
-        cursor: var(--button-cursor--pressed);
+        cursor: var(--button-cursor--pressed, var(--button-cursor, pointer));
       }
 
       &[disabled] {
--- a/ui/src/components/layer/display/layer-display-list-item.ts
+++ b/ui/src/components/layer/display/layer-display-list-item.ts
@@ -343,6 +343,11 @@ export class LayerDisplayListItem extends CoreElement {
       border-color: var(--color-hovered);
     }
 
+    :host(.is-dragged),
+    :host(.is-in-drag) {
+      cursor: grabbing;
+    }
+
     :host > hr {
       --offset-h: 9px;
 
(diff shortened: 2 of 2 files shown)


ui/src/components/core/core-button.ts
```diff
@@ -101,7 +101,7 @@ export class CoreButton extends LitElement {
         color: var(--button-text--pressed);
         background-color: var(--button-bg--pressed);
         border-color: var(--button-border--pressed);
-        cursor: var(--button-cursor--pressed);
+        cursor: var(--button-cursor--pressed, var(--button-cursor, 'pointer'));
```
TIL-EBP, line 104: this seems wrong, did you mean to replace the old value?

ui/src/components/core/core-button.ts
```diff
@@ -76,7 +76,7 @@ export class CoreButton extends LitElement {
       padding: var(--button-padding-v) var(--button-padding-h);
       border: 1px solid;
       border-radius: 4px;
-      cursor: var(--button-cursor, pointer);
+      cursor: var(--button-cursor, pointer) !important;
```
TIL-EBP, line 79: is there a way to do this without important? NO this overrides the behavior for when a layer is disabled

TIL-EBP (changes requested):
There seems to be an error in the updated code and the important keyword breaks the behavior for disabled laers

ui/src/components/core/core-button.ts
daniel-va: No, this is as it should be.
What this means:
- if `--button-cursor--pressed` is set, use that.
- if instead `--button-cursor` is set, use that.
- if none of the two variables is set, use `pointer`.

ui/src/components/core/core-button.ts
daniel-va: I wanted to use this only for debugging purposes - its not needed.

TIL-EBP (approved):
Just a little remark: When releasing a layer while the mouse is hovered above a button from another layer, the first layer does not get dropped but I can perform the action associated with the button when i click it. Also, the cursor changes. Not a deal breaker but might be nice if this could be fixed as well (if possible within a justifiable amount of time)

daniel-va:
@TIL-EBP De drop on a button is probably out-of-scope, timewise. It was already a headache implementing the drag in a way that didn't fully wreck the tooltips/dropdowns.

The cursor changing is due to how Sortable.js handles the display of a currently dragged element. I would leave it like it is, for now, as I don't think fixing it is worth the effort.

PR #1466, opened 2025-02-25
Feature 1374: Struktur Werkzeugpanel // #3 New Panel

daniel-va:
Depends on #1458.

Replaces the tools panel with a new, not-yet-fully-implemented one.


ui/src/elements/ngm-side-bar.ts
```diff
@@ -277,6 +269,14 @@ export class SideBar extends LitElementI18n {
       </div>
     `;
   }
+  //
```
TIL-EBP, line 272: are you still working on this? If not, remove

TIL-EBP (approved):

ui/src/elements/ngm-side-bar.ts
daniel-va: This is removed in my current working tree. I'll leave it in here.

PR #1492, opened 2025-03-24
add Content-Security-Policy header to backend

TIL-EBP:
does not add Content-Security-Header to the front end because some fonts are loaded from cesium


daniel-va (commented):
Why is the CSP only on the backend, and why is the permissions policy only on the frontend?

TIL-EBP:
I did not at CSP in the frontend because we need to load some fonts and other things from Cesium.

daniel-va:
@TIL-EBP We could set the header to something like this:

```
Content-Security-Policy: default-src 'self'; style-src 'self' https://fonts.googleapis.com; style-src 'self' https://some-style-source.cesium.com, ...`
```
I don't know if that is feasible to do, your call.

api/src/lib.rs
```diff
@@ -54,6 +54,14 @@ pub async fn app(pool: PgPool) -> Router {
         .layer(SetResponseHeaderLayer::if_not_present(
             X_FRAME_OPTIONS,
             HeaderValue::from_static("SAMEORIGIN"),
+        ))
+        .layer(SetResponseHeaderLayer::if_not_present(
+            CONTENT_SECURITY_POLICY,
+            HeaderValue::from_static("default-src 'self'"),
+        ))
+        .layer(SetResponseHeaderLayer::if_not_present(
```
daniel-va, line 62: You can add custom headers (see [HeaderName](https://traffloat.github.io/api/master/hyper/header/struct.HeaderName.html)):

```rust
HeaderName::from_static("Permissions-Policy")
```

daniel-va (changes requested):

TIL-EBP:
I couldn't get the CSP to work in the last 15 Minutes so I would say we jut leave it out as it was not required in the first place

daniel-va (approved):

PR #1507, opened 2025-04-01
Fix translated text in dropdowns not being updated

daniel-va:
Fixes an issue where translations in dropdowns was not being updated.
The problem was that the dropdowns were within a draggable element, which, after having been moved, caused the dropdown to loose its reactivity.

Also solves the issue where dropdowns were sometimes losing their content.
This had a similar reason, as Lit removed and re-added some elements containing dropdowns,
which destroyed the contents.

### DIFF
--- a/ui/src/features/core/base/core-base-popup-box.element.ts
+++ b/ui/src/features/core/base/core-base-popup-box.element.ts
@@ -2,6 +2,8 @@ import { css, html, LitElement } from 'lit';
 import { applyTransition } from 'src/styles/theme';
 
 export abstract class CoreBasePopupBox extends LitElement {
+  private timeoutForToggle: unknown = null;
+
   protected constructor() {
     super();
     this.show = this.show.bind(this);
@@ -9,20 +11,28 @@ export abstract class CoreBasePopupBox extends LitElement {
   }
 
   show(): void {
+    this.clearToggle();
     this.classList.remove('is-hidden');
-    setTimeout(() => {
+    this.timeoutForToggle = setTimeout(() => {
       this.classList.add('is-visible');
     });
   }
 
   hide(): void {
+    this.clearToggle();
     this.classList.remove('is-visible');
-    setTimeout(() => {
+    this.timeoutForToggle = setTimeout(() => {
       this.classList.add('is-hidden');
     }, 250);
   }
 
-  readonly render = () => html``;
+  private clearToggle(): void {
+    if (this.timeoutForToggle !== null) {
+      clearTimeout(this.timeoutForToggle as number);
+    }
+  }
+
+  readonly render = () => html`<slot></slot>`;
 
   static readonly styles = css`
     :host,
--- a/ui/src/features/core/base/core-base-popup.element.ts
+++ b/ui/src/features/core/base/core-base-popup.element.ts
@@ -1,21 +1,28 @@
-import { css, html, LitElement } from 'lit';
+import { css, LitElement, render } from 'lit';
 import { property } from 'lit/decorators.js';
 import { CoreBasePopupBox } from 'src/features/core/base/core-base-popup-box.element';
+
+export interface PopupProps {
+  position?: PopupPosition;
+
+  align?: PopupAlign;
+}
 
 export abstract class CoreBasePopup<
   TBox extends CoreBasePopupBox,
 > extends LitElement {
-  @property()
+  @property({ type: String, reflect: true })
   accessor position: PopupPosition | null = null;
 
-  @property()
+  @property({ type: String, reflect: true })
   accessor align: PopupAlign | null = null;
+
+  @property({ type: Object })
+  accessor content: unknown = null;
 
   private _target: Element | null = null;
 
-  protected box: TBox | null = null;
-
-  private isInitialized = false;
+  protected box!: TBox;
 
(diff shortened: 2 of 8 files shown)


ui/src/features/core/base/core-base-popup-box.element.ts
```diff
@@ -2,27 +2,37 @@ import { css, html, LitElement } from 'lit';
...
-    setTimeout(() => {
+    this.timeoutForToggle = setTimeout(() => {
       this.classList.add('is-hidden');
     }, 250);
   }
 
-  readonly render = () => html``;
+  private clearToggle(): void {
+    if (this.timeoutForToggle !== null) {
+      clearTimeout(this.timeoutForToggle as number);
```
Tugark, line 31: I would specify `timeoutForToggle` as `number|null` and then get rid of the explicit cast here. Or is there a reason for doing it this way? As per the docs, `setTimeout` returns a positive integer id.

ui/src/features/core/base/core-base-popup.element.ts
```diff
@@ -74,38 +84,23 @@ export abstract class CoreBasePopup<
...
-    this.box.addEventListener('hide', this.hide);
-
-    this.syncSlot();
-
-    if (this._target == null) {
-      this.registerTarget();
-    }
-    this.isInitialized = true;
+  willUpdate(): void {
+    console.log('renderino!');
```
Tugark, line 100: Should probably be removed

ui/src/features/core/core-dropdown.element.ts
```diff
@@ -77,17 +81,23 @@ export class CoreDropdown extends CoreBasePopup<CoreDropdownBox> {
...
-    <ngm-core-dropdown-box></ngm-core-dropdown-box>
-  `;
-
   static readonly styles = css`
     ${CoreBasePopup.styles}
   `;
 }
+
+export function dropdown(content: unknown): unknown;
+export function dropdown(props: PopupProps, content: unknown): unknown;
```
Tugark, line 90: I guess these items are somehow required by the UI framework, right? Otherwise, they should be removed?

ui/src/features/core/core-tooltip.element.ts
```diff
@@ -23,11 +26,20 @@ export class CoreTooltip extends CoreBasePopup<CoreTooltipBox> {
...
+}
 
-  readonly findBoxElement = () =>
-    this.shadowRoot!.querySelector('ngm-core-tooltip-box')! as CoreTooltipBox;
-
-  readonly renderBox = () => html`
-    <ngm-core-tooltip-box></ngm-core-tooltip-box>
-  `;
+export function tooltip(content: unknown): unknown;
+export function tooltip(props: PopupProps, content: unknown): unknown;
```
Tugark, line 32: Here as well.

ui/src/features/core/core-tooltip.element.ts
```diff
@@ -23,11 +26,20 @@ export class CoreTooltip extends CoreBasePopup<CoreTooltipBox> {
...
+export function tooltip(
+  propsOrContent: unknown,
+  contentOrNone?: unknown,
+): unknown {
+  const content = contentOrNone ?? propsOrContent;
+  const props =
+    contentOrNone === undefined ? {} : (propsOrContent as PopupProps);
```
Tugark, line 39: That one is quite hard to read; but I assume this is another framework specific element, right?
(I mean there's some implicit stuff going on; i.e. that if `contenOrNone` is defined, `propsOrContent` will always be `props` and not `content`).

Tugark (approved):
Some minor remarks; mostly because I'm not familiar with Lit or other frameworks used.

ui/src/features/core/base/core-base-popup-box.element.ts
daniel-va: The typescript config within this project is somehow setup in a way that includes the NodeJS type definitions for `setTimeout`, which doesn't return `number`, but `NodeJS.Timeout`. However, `NodeJS.Timeout` is not available as a type, which is why we have to use this rather awkward typing.

I have this on a long list of things I would like to fix in this project, but for now, it will probably have to stay.

ui/src/features/core/core-dropdown.element.ts
daniel-va: These are just overloads to make the use of `dropdown` safer. Without them, the first argument would never have type checking while a second one is specified.

ui/src/features/core/core-tooltip.element.ts
daniel-va: See above.

ui/src/features/core/core-tooltip.element.ts
daniel-va: That's just TypeScript overload stuff - don't know how familiar you are with these, but this is rather tame 😋

daniel-va: (To add, prettier does to an excellent job at creating unreadable code here)

Tugark: Yeah, if you read it with the overloads you mentioned, it makes sense. Other than that, it was ugly (even though it turned out the same :D)

PR #1518, opened 2025-04-03
Feature/viewer 1482 datenkatalog design

### LINKED ISSUE #1482: Datenkatalog | Design implementieren
**Beschreibung**
Das Design des Datenkatalogs im Daten-Panel ist noch nicht gemäss  https://www.figma.com/design/cEiOoOazAQZqpRY92ZhBeO/swissgeol?node-id=9802-7247&t=oGa5WFuJdAp3p9uf-0 implementiert.

**Design**
- [ ] Design implementieren

TIL-EBP:
resolves #1482

### DIFF
--- a/ui/src/features/core/core-accordion.element.ts
+++ b/ui/src/features/core/core-accordion.element.ts
@@ -1,0 +1,49 @@
+// my-accordion.ts
+import { css, html, LitElement } from 'lit';
+import { customElement, query, state } from 'lit/decorators.js';
+
+@customElement('ngm-core-accordion')
+export class CoreAccordion extends LitElement {
+  @state() private accessor isOpen = false;
+  @query('slot[name="header"]') private accessor headerSlot!: HTMLSlotElement;
+
+  private toggle() {
+    this.isOpen = !this.isOpen;
+    this.updateHeaderClass();
+  }
+
+  private updateHeaderClass() {
+    const headerNodes =
+      this.headerSlot?.assignedElements({ flatten: true }) ?? [];
+    headerNodes.forEach((node) => {
+      if (node instanceof HTMLElement) {
+        node.classList.toggle('active', this.isOpen);
+      }
+    });
+  }
+
+  render() {
+    return html`
+      <div class="header" @click=${this.toggle}>
+        <slot name="header"></slot>
+      </div>
+      <div class="content" ?open=${this.isOpen}>
+        <slot></slot>
+      </div>
+    `;
+  }
+
+  static readonly styles = css`
+    .header {
+      cursor: pointer;
+    }
+
+    .content {
+      display: none;
+    }
+
+    .content[open] {
+      display: block;
+    }
+  `;
+}
--- a/ui/src/features/layer/layer-catalog-category.element.ts
+++ b/ui/src/features/layer/layer-catalog-category.element.ts
@@ -1,0 +1,149 @@
+import { customElement, property, state } from 'lit/decorators.js';
+import { CoreElement } from 'src/features/core';
+import { type LayerTreeNode } from 'src/layertree';
+import { css, html, TemplateResult } from 'lit';
+import i18next from 'i18next';
+import { repeat } from 'lit/directives/repeat.js';
+import { getCategoryOrLayerTemplate } from 'src/features/layer/layer-catalog.element';
+import auth from 'src/store/auth';
+
+@customElement('ngm-layer-catalog-category')
+export class LayerCatalogCategory extends CoreElement {
+  @property() accessor node: LayerTreeNode | null = null;
+  @property() accessor isTopLevel = false;
+
+  @state()
+  private accessor userGroups: string[] = [];
+
+  connectedCallback() {
+    super.connectedCallback();
+    this.register(
+      auth.user.subscribe((user) => {
+        this.userGroups = user?.['cognito:groups'] ?? [];
+      }),
+    );
+  }
+
+  findNumberOfActiveLayersForCategory(layer: LayerTreeNode): number {
+    const children = layer.children;
+    if (children == null) {
+      return 0;
+    }
+    let active_layers = 0;
+    children.forEach((child) => {
+      if (child.children) {
+        active_layers += this.findNumberOfActiveLayersForCategory(child);
+      }
+      if (child.displayed) {
+        active_layers++;
+      }
+    });
+    return active_layers;
+  }
+
+  render() {
+    if (this.node == null) {
+      return html``;
+    }
+    // if it is a restricted layer, the user must be logged in to see it
+    const children = this.node.children?.filter(
+      (node) =>
+        !(
+          node.restricted &&
+          !node.restricted.some((g) => this.userGroups.includes(g))
+        ),
+    );
+    if (children == null || children.length === 0) {
+      return html``;
+    }
+    let header: TemplateResult;
+    if (this.isTopLevel) {
+      const activeLayers = this.findNumberOfActiveLayersForCategory(this.node);
+      header = html`
+        <div class="header-title">
+          <span>${i18next.t(this.node.label)}</span>
+          ${activeLayers > 0
+            ? html`<ngm-core-chip>${activeLayers}</ngm-core-chip>`
+            : html``}
+        </div>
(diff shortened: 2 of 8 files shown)


ui/src/features/layer/layer-catalog.element.ts
```diff
@@ -39,15 +39,35 @@ export class LayerCatalog extends CoreElement {
```
daniel-va, line 39: Unrelated to this PR, but I noticed that this component adds a subscription in its constructor that is never unsubscribed. That should probably be moved into `connectedCallback`, where it can be handled via `register(...)`.

ui/src/features/layer/layer-catalog.element.ts
```diff
@@ -59,17 +79,36 @@ export class LayerCatalog extends CoreElement {
...
+        <ngm-core-icon icon="dropdown"></ngm-core-icon>
+      `;
+    } else {
+      header = html`
+        <ngm-core-icon icon="dropdown"></ngm-core-icon>
+        <label>${i18next.t(category.label)}</label>
+      `;
+    }
     return html`
       <div class="category ui accordion">
```
daniel-va, line 101: I would really like to get rid of this accordion. Do you have the capacity to refactor this into a core component in this issue?

ui/src/features/layer/layer-catalog.element.ts
```diff
@@ -110,131 +145,96 @@ export class LayerCatalog extends CoreElement {
     html`${repeat(
       this.layers,
       (node) => node,
```
daniel-va, line 147: This would probably be a better key, in case the layer changes:
```suggestion
      (node) => node.layer ?? node.label ?? node,
```
Layers are definitely missing a required, unique id.

ui/src/features/layer/layer-catalog.element.ts
```diff
@@ -110,131 +145,96 @@ export class LayerCatalog extends CoreElement {
...
+    .ui.accordion .title.top-level-header {
+      display: flex;
+      justify-content: space-between;
+      align-items: center;
+      padding: 12px 16px;
 
-    .ngm-checkbox input:checked + .ngm-checkbox-icon {
-      background-color: var(--color-action);
+      &.active {
+        border-bottom: 2px solid var(--color-bg--dark);
```
daniel-va, line 216: This is an outdated variable.

ui/src/features/layer/layer-catalog.element.ts
```diff
@@ -39,15 +39,35 @@ export class LayerCatalog extends CoreElement {
...
   ): TemplateResult {
     if (node.children) {
-      return this.getCategoryTemplate(node, level);
+      return this.getCategoryTemplate(node, isTopLevel);
     }
     return this.getLayerTemplate(node);
   }
 
-  getCategoryTemplate(category: LayerTreeNode, level: string): TemplateResult {
+  findNumberOfActiveLayersForCategory(layer: LayerTreeNode): number {
```
daniel-va, line 50: This function is currently called for every layer on every render. Maybe move it into a component where we can control when the entire thing has to be re-computed?

ui/src/features/layer/layer-catalog.element.ts
```diff
@@ -39,15 +39,35 @@ export class LayerCatalog extends CoreElement {
...
+        active_layers += this.findNumberOfActiveLayersForCategory(child);
+      }
+      if (child.displayed) {
+        active_layers++;
+      }
+    });
+    return active_layers;
+  }
+
+  getCategoryTemplate(
```
daniel-va, line 67: I'd prefer if we move this into its own component. Its quite complex and gets called often.

daniel-va (changes requested):
Looks nice UI-wise :) 
There are some things I'd like to see cleaned up in the code, but please judge for yourself which ones you find to be necessary.

ui/src/features/layer/layer-catalog.element.ts
TIL-EBP: I have started working on a core-accordion, but I am struggeling with the nested accordions

ui/src/features/layer/layer-catalog.element.ts
TIL-EBP: I am not quite sure how we can avoid this with the current structure. Maybe lets talk about this next week

ui/src/features/core/core-accordion.element.ts
```diff
@@ -0,0 +1,47 @@
+// my-accordion.ts
+import { css, html, LitElement } from 'lit';
+import { customElement, query, state } from 'lit/decorators.js';
+
+@customElement('ngm-core-accordion')
+export class CoreAccordion extends LitElement {
+  @state() private accessor isOpen: boolean = false;
```
daniel-va, line 7:
```suggestion
  @state() private accessor isOpen = false;
```

ui/src/features/core/core-accordion.element.ts
```diff
@@ -0,0 +1,47 @@
...
+    this.isOpen = !this.isOpen;
+    this.updateHeaderClass();
+  }
+
+  private updateHeaderClass() {
+    const headerNodes =
+      this.headerSlot?.assignedElements({ flatten: true }) ?? [];
+    headerNodes.forEach((node) => {
+      if (node instanceof HTMLElement) {
+        node.classList.toggle('active', this.isOpen);
```
daniel-va, line 20: Did we ever define how we are naming css classes? If not, then I'd be up for writing down a standard somewhere.

ui/src/features/layer/category-template.element.ts
```diff
@@ -0,0 +1,155 @@
...
+import { type LayerTreeNode } from 'src/layertree';
+import { css, html, TemplateResult } from 'lit';
+import i18next from 'i18next';
+import { repeat } from 'lit/directives/repeat.js';
+import { getCategoryOrLayerTemplate } from 'src/features/layer/layer-catalog.element';
+import { Subscription } from 'rxjs';
+import auth from 'src/store/auth';
+
+@customElement('ngm-category-template')
+export class CategoryTemplate extends CoreElement {
```
daniel-va, line 12: Calling this `Template` seems confusing to me. IMO something like `LayerCatalogGroup` would make more sense.

ui/src/features/layer/category-template.element.ts
```diff
@@ -0,0 +1,155 @@
...
+  @property() accessor node!: LayerTreeNode;
```
daniel-va, line 13: Just a heads up that in Lit, there are no required properties, so marking this as `node!` doesn't really do anything. The "safe" way to do this would be to make it `node: LayerTreeNode | null = null` and then handle the absence explicitly.

ui/src/features/layer/category-template.element.ts
```diff
@@ -0,0 +1,155 @@
...
+  @property() accessor isTopLevel: boolean = false;
```
daniel-va, line 14:
```suggestion
  @property() accessor isTopLevel = false;
```

ui/src/features/layer/category-template.element.ts
```diff
@@ -0,0 +1,155 @@
...
+
+  @state()
+  accessor userGroups: string[] = [];
```
daniel-va, line 17:
```suggestion
  private accessor userGroups: string[] = [];
```

daniel-va (changes requested):
Looks good overall, I have some small issues with naming/typing remaining.

ui/src/features/core/core-accordion.element.ts
TIL-EBP: I don't think we did, but would be down to do so

ui/src/features/layer/layer-catalog-category.element.ts
```diff
@@ -28,7 +26,6 @@ export class CategoryTemplate extends CoreElement {
 
   disconnectedCallback() {
```
daniel-va, line 27: Can be removed.

daniel-va (approved):
LGTM :)

Seems like you will need to recreate this branch due to the repo renaming of yesterday - the PR currently contains 92 commits...

PR #1540, opened 2025-05-12
Feature/viewer 1430 cesium ion upload

### LINKED ISSUE #1430: Design für Cesium-ion upload erstellen und implementieren
**Beschreibung**
Im Upload Tap können über den Button "Inhalte von Cesium ion hinzufügen" Inhalte von Cesium Ion hochgeladen werden.
für die jeweiligen Floating-Fenster (siehe nachfolgenden Screenshots) besteht derzeit noch kein Design. Dies soll in diesem Issues erstellt werden.

Image

Image

**ToDo**
- [ ] Design für die oben abgebildeten Floating-Fenster erstellen.

TIL-EBP:
resolves #1430


ui/src/elements/ngm-ion-modal.ts
```diff
@@ -155,7 +238,154 @@ export class NgmIonModal extends LitElementI18n {
           </table>
         </div>
       </div>
-      ${dragArea}
+      <hr />
+      <div class="actions">
+        <sgc-button
+          color="tertiary"
+          role="link"
```
daniel-va, line 245: Instead of using `role="link"` and opening the page via JS, we should probably consider adjusting the `sgc-button` component so it supports links - I'm fine leaving this as-is for now, but this is not the cleanest way to handle this in general.

ui/src/elements/ngm-ion-modal.ts
```diff
@@ -155,7 +238,154 @@ export class NgmIonModal extends LitElementI18n {
...
+        ></sgc-button>
+        <sgc-button color="secondary" @buttonClick=${() => this.onClose()}
+          >${i18next.t('app_close_btn_label')}</sgc-button
+        >
+      </div>
     `;
   }
+
+  static readonly styles = css`
+    :host {
```
daniel-va, line 259: Might be a good idea to configure `box-sizing` here to ensure your sizes work correctly.
```css
    :host,
    :host * {
      box-sizing: border-box;
    }
```

ui/src/elements/ngm-ion-modal.ts
```diff
@@ -155,7 +238,154 @@ export class NgmIonModal extends LitElementI18n {
...
+    .content,
+    .actions {
+      padding: 24px;
+    }
+
+    hr {
+      margin: 0;
+      height: 1px;
+      border-width: 0;
+      color: var(--color-border--default);
```
daniel-va, line 274: We probably should start transitioning to using the variables from `swissgeol-ui-core`.

ui/src/elements/ngm-ion-modal.ts
```diff
@@ -155,7 +238,154 @@ export class NgmIonModal extends LitElementI18n {
...
+
+    .actions {
+      display: flex;
+      justify-content: flex-end;
+      gap: 16px;
+      margin-top: 8px;
+    }
+
+    sgc-checkbox {
+      box-sizing: border-box;
```
daniel-va, line 324: If this isn't the default for `sgc-checkbox` already, then we definitely should add this to the core library.

ui/src/features/core/core-button.element.ts
```diff
@@ -213,6 +213,11 @@ export class CoreButton extends LitElement {
       padding: var(--button-padding-v);
     }
 
+    /* small shape */
+    :host([shape='small']) button {
+      padding: 2px 8px;
```
daniel-va, line 218: Can you adjust `--button-padding-v` and `-h` instead of directly setting the padding?

ui/src/features/core/core-button.element.ts
```diff
@@ -236,6 +241,12 @@ export class CoreButton extends LitElement {
 
 export type ButtonVariant = 'primary' | 'secondary' | 'tertiary';
 
-export type ButtonShape = 'default' | 'large' | 'icon' | 'icon-round' | 'chip';
+export type ButtonShape =
+  | 'default'
+  | 'large'
+  | 'small'
```
daniel-va, line 247: Should we also add this new shape to the core library?

ui/src/features/core/core-checkbox.element.ts
```diff
@@ -29,20 +32,40 @@ export class CoreCheckbox extends LitElement {
...
-          viewBox="0 0 12 10"
-          fill="none"
-          xmlns="http://www.w3.org/2000/svg"
-        >
-          <path
-            d="M11.2585 0.990983C11.4929 1.22539 11.6245 1.54328 11.6245 1.87473C11.6245 2.20619 11.4929 2.52407 11.2585 2.75848L5.00853 9.00848C4.77412 9.24282 4.45623 9.37447 4.12478 9.37447C3.79332 9.37447 3.47544 9.24282 3.24103 9.00848L0.741026 6.50848C0.513329 6.27273 0.387336 5.95698 0.390184 5.62923C0.393032 5.30149 0.524493 4.98797 0.756253 4.75621C0.988014 4.52445 1.30153 4.39299 1.62927 4.39014C1.95702 4.38729 2.27277 4.51329 2.50853 4.74098L4.12478 6.35723L9.49103 0.990983C9.72544 0.756644 10.0433 0.625 10.3748 0.625C10.7062 0.625 11.0241 0.756644 11.2585 0.990983Z"
-            fill="currentColor"
-          />
-        </svg>
+        ${this.isIndeterminate
```
daniel-va, line 40: Is this actually used by you in this PR? Do you now not just use use the core library?

ui/src/features/core/core-modal.element.ts
```diff
@@ -9,6 +9,11 @@ interface CoreModalProps {
    */
   isPersistent?: boolean;
   hasNoPadding?: boolean;
+  /**
+   * When using the modal with sgc components, there is a weird behavior that adds a scroll bar on the entire dialog if the dialog content already has one.
```
daniel-va, line 13: While I like adding a comment here to document what this attribute does, I do not think we need to detail why it has been added - as far as I can see, the problem lies not inherently with the core library, but rather some weird CSS behavior that could also happen with any other component. Also, its not like every core component causes this.

I personally would just change this comment to say something along the lines of `this attribute hides the overflow of the modal` without further clarification as to why it has been added.

ui/src/features/core/core-text-input.element.ts
```diff
@@ -0,0 +1,94 @@
...
+  accessor label: string = '';
+
+  @property()
+  accessor placeholder: string = '';
+
+  @property()
+  accessor value: string = '';
+
+  @property({ type: Boolean, attribute: 'is-search', reflect: true })
+  accessor search: boolean = false;
```
daniel-va, line 17: I would prefer having a generic `icon` property over this specific `search` boolean.

ui/src/index.ts
```diff
@@ -1,8 +1,10 @@
...
 import './style/index.css';
 import { ReactiveElement } from 'lit';
+import { defineCustomElements } from '@swisstopo/swissgeol-ui-core/loader';
+import './ngm-app-boot';
 
 // Detect issues following lit2 migration
 ReactiveElement.enableWarning?.('migration');
 
-import './ngm-app-boot';
+defineCustomElements();
```
daniel-va, line 10: Consider calling `defineCustomElements` before importing `ngm-app-boot`, just so that the core elements are ready when the app is initializing.

daniel-va (changes requested):

ui/src/elements/ngm-ion-modal.ts
TIL-EBP: Yes, this is a good point. In general, how should we proceed when using components from core that are missing behavior or styles?

daniel-va: There is a project for core already: https://github.com/orgs/swisstopo/projects/18
So we could maybe start adding feature requests there?

ui/src/elements/ngm-ion-modal.ts
TIL-EBP: added a ticket

daniel-va (approved):
Looking good - don't forget to squash.

PR #1555, opened 2025-05-19
Feature 1465: Integration swissBEDROCK

### LINKED ISSUE #1465: Integration swissBEDROCK
**Beschreibung**
 Daten Datensatz `swissBEDROCK` soll in den Viewer integriert werden . Folgende Schritten sind dafür geplant:

- [ ] Nur GeoTiff (12 Bänder) als ein Layer (bestehenden Layer in `Geometrie des Untergrunds > Felsoberfläche > Feldsoberfläche` ersetzen) integrieren. Per Mouse-click sollen die werte sämtlicher Bänder in einem Floating-Panel (vergleiche Voxeldaten) angezeigt werden.
- [ ] Design für Wechselfunktion für das angezeigt Band erstellen. Eine kombinierte Filterfunktion (vergleiche Voxelfilter) sämtlicher Bänder mit in das Design einbeziehen und in Figma dokumentieren.
- [ ] Wechselfunktion der Bänder gemäss Figma implementieren
- [ ] Felsoberfläche als 3D-Horizont oder Höhenmodell integrieren
- [ ] Kombinierte Filterfunktion gemäss Figma implementieren

daniel-va:
Resolves #1465.

**Don't merge this until a release candidate for 1.1.0 has been released!**

Adds the new *swissBEDROCK* layer under *Geometrie des Untergrunds > Felsoberfläche*.
- The layer's active band can be selected via the filters menu.
- Data about a point on the layer can be displayed by right-clicking anywhere on the layer.

To render the geoTIFF, the new [`TiTiler`](https://developmentseed.org/titiler/) service has been added. That service requires a custom Docker image, which will now be built automatically on each deploy. When first deploying this PR to a K8s cluster, the [deploy-k8s](https://github.com/swisstopo/swissgeol-viewer-suite/actions/workflows/deploy-k8s.yml) needs to be run to make that service available.

Note that the performance of the layer can be quite bad when running it locally as the TiTiler service has rather low resource bounds so it doesn't eat the entire machine's memory and CPU. On K8s, the service has been limited to 1 CPU and 5GiBs of memory. After deploying, we should monitor how those limits perform and adjust them accordingly.


ui/src/features/core/core-button.element.ts
```diff
@@ -207,6 +207,10 @@ export class CoreButton extends LitElement {
       --button-bg: transparent;
     }
 
+    :host([transparent][variant='tertiary']) {
+      --button-border: transparent;
```
TIL-EBP, line 211: Is this already implemented in core? if not, maybe add a ticket for this.

TIL-EBP (approved):
Looks and works nice. 
Only a few things: 

Clicking on the map for detail is a little wonky and the "zoom to" feature does not really zoom to wehre I clicked/ changes camera orientation and i very quickly lose the green point . This is only an issue when doing these things in 3D view (as you mentioned). So I suggest we leave it as is and wait for their feedabck

PR #1556, opened 2025-05-19
Feature #1548: Voxel-Filter Panel schliessen - Filter-Status beihalten

### LINKED ISSUE #1548: Voxel-Filter Panel schliessen - Filter-Status beihalten
**Beschreibung**
Bei schliessen des Voxel-Filter Panels sollen die Fitlereinstellungen beibehalten werden. Um den Filter zu deaktivieren muss der Filter neu aus dem Kontextmenu aufgerufen werden.

**Abgrenzung**

**Acceptance Criteria**

- [x] Die Antwort muss 42 sein.
- [ ] ..Filter wird bei Schliessen nicht mehr deaktiviert

daniel-va:
Resolves #1548.

Persists voxel filters across multiple dialogs by reading their state from the voxel shader.


ui/src/elements/ngm-voxel-filter.ts
```diff
@@ -60,55 +105,48 @@ export class NgmVoxelFilter extends LitElementI18n {
 
   close() {
     this.hidden = true;
-    this.resetShader();
+    this.isLithologyActiveByIndex = [];
```
TIL-EBP, line 108: this also needs to happen when the filter is closed because a new one was opened while the old was still open.
Currently, when I open a filter while one is stil open, it uses the om the old one instead of its own values

daniel-va: Fixed

PR #1581, opened 2025-05-26
Feature 1414: 2D view Fix navigation

### LINKED ISSUE #1414: 2D view Fix navigation
Navigation fixieren, so dass man nur folgende Navigationsmöglichkeiten hat:
- Neigung: -90° (fixiert)
- Pan
- Zoom

Zusätzlich wird die Terrain Durchdringung (`viewer.scene.globe.depthTestAgainstTerrain = true;
viewer.scene.screenSpaceCameraController.enableCollisionDetection = true;`) ausgeschaltet, so dass man mit der Kamera nicht unter die Erde fahren kann.

Wenn Kamera initial unter Terrain, soll die Kamera über das Terrain bewegt werden. Auf Schweiz zoom. 

https://www.figma.com/design/cEiOoOazAQZqpRY92ZhBeO/swissgeol?node-id=13540-72552&t=Yasz5aMRtI3I0ldP-4

Abgrenzung:
- UI wird nicht angepasst / nur 2D Button hinzufügen, kein Redesign

daniel-va:
Resolves #1414.

Adds a new navigation control that, when enabled, locks the camera in a top-down view, preventing the user modifying camera tilt.


TIL-EBP (changes requested):
Looks good, but i think you forgot about theis: 
image

And one minor thing: 
When 2D view is enabled, there is no way of rotating the map (Azimuth) by mouse (or I don't know how), but the Azimuth control is still enabled in the cam config. NOt sure what the desired behavor is there


I just played around a little more and when pressing SHIFT, then the 2D only mode is overriden and i can still change the pitch

TIL-EBP (approved):
When below the surface when activating the "D mode, the map should zoom to switzerland according to the ticket descripton. Currently it zooms to almost ground level

PR #1645, opened 2025-08-06
Bugfix 1634: Layersortierung funktioniert nicht korrekt

### LINKED ISSUE #1634: Layersortierung funktioniert nicht korrekt
**Beschreibung**

Die Layersortierung funktioniert gelegentlich (schwer reproduzierbar) nicht.
Sobald nur Layer aus Datenkatalog über die Suche hinzugefügt wurden, tritt der Fehler auf

**Abgrenzung**

**Acceptance Criteria**

- [ ] Die Antwort muss 42 sein.
- [ ] ...

daniel-va:
Resolves #1634.

Refactors the layer catalog and display list to use `LayerService` for all layer activation and deactivation. This is meant to solve multiple issues with layer handling, such as failing to reorder, remove or syncing url and catalog.

Note that I was no longer able to reproduce the sorting bug (or any other issue with the active layers). However, I am not fully certain that it has been eliminated. This PR just contains changes to hopefully make the whole layer handling a lot more robust and thus eliminate unexpected behavior altogether.

### DIFF
--- a/ui/src/elements/ngm-side-bar.ts
+++ b/ui/src/elements/ngm-side-bar.ts
@@ -176,7 +176,7 @@ export class SideBar extends LitElementI18n {
         label: asset.name,
         layer: asset.id.toString(),
         visible: true,
-        displayed: true,
+        displayed: false,
         opacityDisabled: true,
         pickable: true,
         customAsset: true,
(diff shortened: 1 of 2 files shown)


TIL-EBP (changes requested):
Some things i found while comparing to DEV: 

- The layerCount does not update correctly (updates only after opening and closing the data catalogue
- Removing layers like "Felsoberfläche" does no longer work (Removing from the layed layer works better than on DEV, but not from the actual View)
- I can add the same LAyer from the Search multiple times. Removing one of them leads to weird behavior (Removing the first causes it to be set to invisible, removing it again removes it. The second one can no longer be removed then, but when i try to reorder it, it disapears)
- Some layers (like Tiefe Bohrungen) are not displayed on the map when added via search
-

PR #1750, opened 2025-11-06
Color new 3d tiles according to property colors and display custom properties

### LINKED ISSUE #1707: 3dtiles | Neue Properties auslesen
**Beschreibung**
Properties aus neuem Ort in der Klasse auslesen
Farbe aus Properties auslesen

siehe #1665 und #1622 

**Abgrenzung**

**Acceptance Criteria**

- [ ] ...

TIL-EBP:
resolves #1707

### DIFF
--- a/ui/package-lock.json
+++ b/ui/package-lock.json
@@ -1,0 +1,18790 @@
+{
+  "name": "ngm",
+  "lockfileVersion": 3,
+  "requires": true,
+  "packages": {
+    "": {
+      "name": "ngm",
+      "dependencies": {
+        "@aws-sdk/client-cognito-identity": "3.768.0",
+        "@aws-sdk/client-s3": "3.758.0",
+        "@aws-sdk/credential-provider-cognito-identity": "3.768.0",
+        "@aws-sdk/s3-request-presigner": "3.758.0",
+        "@geoblocks/ga-search": "0.0.22",
+        "@lit/context": "^1.1.3",
+        "@lit/task": "^1.0.1",
+        "@swissgeol/ui-core": "1.6.0-dev9",
+        "cesium": "1.135.0",
+        "d3-array": "3.2.4",
+        "d3-axis": "3.0.0",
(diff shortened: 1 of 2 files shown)


ui/src/features/layer/info/pickers/layer-info-picker-for-3dtiles.ts
```diff
@@ -158,10 +191,27 @@ class LayerInfoFor3dTile implements LayerInfo {
...
+        .split(', ')
+        .at(-1)
+        ?.slice(0, -1);
+      this.originalColor = new Color(
+        this.feature.customProperties.color_red as number,
+        this.feature.customProperties.color_green as number,
+        this.feature.customProperties.color_blue as number,
+        alpha as unknown as number,
+      );
+      const color = `color("rgb(${OBJECT_HIGHLIGHT_COLOR.red * 255}, ${OBJECT_HIGHLIGHT_COLOR.green * 255}, ${OBJECT_HIGHLIGHT_COLOR.blue * 255})", 1)`;
```
daniel-va: This can be written a tad cleaner with `vec4`:
```suggestion
      const color = `vec4(${OBJECT_HIGHLIGHT_COLOR.red}, ${OBJECT_HIGHLIGHT_COLOR.green}, ${OBJECT_HIGHLIGHT_COLOR.blue}), 1)`;
```
I'm assuming that `color` is passed directly to the shader, for which I'd assume `vec4` to be just a bit faster and/or cleaner than having to parse an rgb color string.

ui/src/features/layer/info/layer-info-item.element.ts
```diff
@@ -105,6 +112,9 @@ export class LayerInfoItem extends CoreElement {
                 }
                 return i18next.t(it.key);
               });
+              if (PROPERTIES_TO_EXCLUDE.includes(key)) {
```
daniel-va: Why filter these out here instead of within `extractTilesetAttributes`?

PR #1843, opened 2026-02-12
Feature/viewer integrate new voxel data

nOester:
Hallo Till wie besprochen hier die  neuen Voxeldaten.


TIL-EBP (approved):
Note: The Cypress tests currently fail, I created an [issue](https://github.com/swisstopo/swissgeol-assets-suite/issues/867) to fix them

PR #1911, opened 2026-04-21
projects switched off

nOester:
@jannic Projects switched off
#1909

### DIFF
--- a/ui/src/features/layout/sidebar/layout-sidebar.element.ts
+++ b/ui/src/features/layout/sidebar/layout-sidebar.element.ts
@@ -121,10 +121,12 @@ export class LayoutSidebar extends CoreElement {
         icon: 'tools',
         counter: this.countOfGeometries,
       })}
-      ${this.renderItem({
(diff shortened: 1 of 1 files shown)


ui/src/features/layout/sidebar/layout-sidebar.element.ts
```diff
@@ -121,10 +121,12 @@ export class LayoutSidebar extends CoreElement {
         icon: 'tools',
         counter: this.countOfGeometries,
       })}
-      ${this.renderItem({
+      <!-- Temporarilly switched off
```
vej-ananas: @nOester commenting does not work here. just delete the lines.

vej-ananas (changes requested):

ui/src/features/layout/sidebar/layout-sidebar.element.ts
nOester: Lines deleted

vej-ananas (approved):

PR #1912, opened 2026-04-22
Ensure that values visibility is correctly read from the url

### LINKED ISSUE #1897: Bug: URL-Parameter | "layers_visibility" funktioniert nicht
**Beschreibung**
Der URL-Parameter `layers_visibility` wird nicht korrekt übergeben, so dass er beim Einfügen der URL in den Browser oder bei der Nutzung eines Permalinks nicht beachtetet wird. Dadurch werden immer alle Layer, auch die ursprünglich ausgeblendeten, angezeigt.

Beispiel:
URL: https://viewer.swissgeol.ch/?lang=en&lon=6.32947&lat=44.50518&elevation=163248&heading=13&pitch=-32&map_transparency=0.65&zExaggeration=2&layers=ch.swisstopo.swissbedrock-geotiff-with-terrain,ch.swisstopo.geologie-geocover&layers_transparency=0,0&layers_visibility=true,false

Permalink: https://link.swissgeol.ch/6clkk 

**Schritte zum Reproduzieren**
1. URL (s. oben in Beschreibung) anklicken (oder in Browser kopieren) oder Permalink anklicken
2. Beide Layer werden angezeigt, obwohl die `layers_visibility` für den zweiten `ch.swisstopo.geologie-geocover` auf `false` steht.

**Beobachtetes Verhalten**
siehe oben

**Erwartetes Verhalten**
Die Layer werden gemäss ihrer `layers_visibility` angezeigt.

vej-ananas:
resolves #1897

### DIFF
--- a/ui/src/utils/type.utils.ts
+++ b/ui/src/utils/type.utils.ts
@@ -1,0 +1,10 @@
+export function parseBoolean(value: string, defaultValue = true): boolean {
+  switch (value.toLowerCase()) {
+    case 'true':
+      return true;
+    case 'false':
+      return false;
+    default:
+      return defaultValue;
+  }
+}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,13 +4,21 @@
 
 ### Added
 
+### Changed
+
+- The project tab is no longer visible until the project feature is fully refactored and ready to be used.
+
+### Fixed
+
+- Layer visibility is now correctly loaded from the url parameters on page reload or when sharing a link.
+
+## [v1.11.0]
+
+### Added
+
 - Infobox for Jura3D layers added
 - 2D-Seismik Survey Geneva added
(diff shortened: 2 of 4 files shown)


ui/src/utils/type.utils.ts
```diff
@@ -0,0 +1,10 @@
+export function parseBoolean(value: string, defaultValue = true): boolean {
```
wendlans, line 1: Why not return `boolean | undefined` and let the caller decide what to do with that information?

wendlans (approved):
I've added just one comment. Otherwise everything looks fine - therefore the approval.

ui/src/utils/type.utils.ts
vej-ananas: Like this, we define the same default behavior for all users of the function. This makes it harder to create different default scenarios out of ignorance.

PR #1915, opened 2026-04-24
Feature/viewer 1914 integrates sa3 d

nOester:
@vej-ananas 
SwissAlps3D layer added including infobox
#1914


vej-ananas (approved):
looks good to me.

Does the cognito group `viewer.swissgeol-dev-sa3d-privileged` exist yet? It seems I can access the data without logging in.

PR #1926, opened 2026-04-28
Add integration test for ogc stac api

vej-ananas:
#1877


.github/copilot-instructions.md
Dead0utside, line 1: Sollen wir das in das Remote-Repo aufnehmen?

Dead0utside (dismissed):

Dead0utside:
Es wäre gut, die Sonar Bemerkungen zu implementieren, sind aber nicht kritisch

ui/src/features/ogc/ogc.service.ogc.test.ts
```diff
@@ -0,0 +1,397 @@
...
+        console.warn(`  [Cleanup] Failed to delete job ${job.id}:`, error);
+      }
+    }
+    createdJobs.length = 0;
+
+    // Clean up temp directory
+    const tempDir = join(process.cwd(), 'temp-ogc-tests');
+    if (existsSync(tempDir)) {
+      try {
+        // await rm(tempDir, { recursive: true, force: true });
```
Dead0utside: shouldn't the temporary files be removed in the end?

ui/src/features/session/session.service.ts
```diff
@@ -122,9 +123,9 @@ export class SessionService extends BaseService {
 
   signIn(): void {
     const url =
-      `https://ngm-${COGNITO_VARIABLES.env}.auth.eu-west-1.amazoncognito.com/oauth2/authorize?` +
+      `https://ngm-${this.cognitoVariables(window.location.host).env}.auth.eu-west-1.amazoncognito.com/oauth2/authorize?` +
```
Dead0utside: `this.cognitoVariables` is initialized as a value, not a function. The compilation might fail here

.github/copilot-instructions.md
vej-ananas: Grundsätzlich ist dasd definitiv ein file, das wir in das Repo aufnehmen wollen. Wir können das aber auch diskutieren.

ui/src/features/ogc/ogc.service.ogc.test.ts
vej-ananas: currently not, since currently the point of this test is to download the files and manually review them. I will change this though so that there is no commented out code but rather a flag which controls this.

ui/src/features/ogc/ogc.service.ogc.test.ts
```diff
@@ -0,0 +1,398 @@
...
+    );
+    expect(res.status).toBe(200);
+
+    const blob = await res.blob();
+    expect(blob.size).toBeGreaterThan(0);
+    console.log(`  [${job.id}] Downloaded ${blob.size} bytes`);
+
+    // Create temporary directory: collection-timestamp-jobId
+    const now = new Date();
+    const timestamp = [
```
uhon, line 205: I personally find it hard to read the time value with all parts separated with dashes.
Also following `job.id` can be confused with milliseconds.

Maybe ISO-String (BUT filename-safe) would be better readable
```
new Date()
  .toISOString()
  .replace(/[:.]/g, '-');
```
If more customized formats are required, i'd go with [Intl.DateTimeFormat](https://developer.mozilla.org/de/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat)

Just a matter of taste tough

.github/copilot-instructions.md
```diff
@@ -0,0 +1,139 @@
+# Copilot Instructions — swissgeol-viewer-app
```
uhon, line 1: Not part of the ticket. Is it accidentally committed or a helpful improvement for future work with Copilot?

ui/src/features/ogc/ogc.http
```diff
@@ -0,0 +1,5 @@
+### GET job status
```
uhon, line 1: What is this?

ui/src/features/ogc/ogc.service.ogc.test.ts
```diff
@@ -0,0 +1,398 @@
...
+ * 1. Submit a job to the OGC API
+ * 2. Poll until job completes
+ * 3. Download the result file
+ * 4. Extract files into a folder prefixed by the ogcSource collection
+ * 5. Check for error files (stac@swisstopo_error.txt) — fail if any are found
+ * 6. Clean up the job and temp files
+ */
+
+const ERROR_FILE_NAME = 'stac@swisstopo_error.txt';
+const shouldDeleteFilesAfterTests = false;
```
uhon, line 108: Maybe this could be made configurable with an ENV Variable.

```
const shouldDeleteFilesAfterTests = process.env.OGC_CLEANUP_FILES === 'true';
```

ui/public/images/compass.gltf
```diff
@@ -5,291 +5,107 @@
   },
```
uhon, line 5: Better readable for sure, but not part of the ticket.
If code re-formating is still desired, we should configure prettier so that it handles `gltf` and `json` format the same everywhere. Maybe even with a pre-commit hook as in assets.

uhon (changes requested):
Just a few minor issues and questions.

ui/src/features/ogc/ogc.service.ogc.test.ts
uhon: no priority

uhon (approved):
All comments non-critical

PR #1950, opened 2026-05-13
#1918: 3D Navigationsanzeige

Dead0utside:
Die Kompassmodelle sieht momentan nicht ganz richtig aus, aber die wird ersetzt sofern es eine korrekte modelle gibt


ui/package.json
```diff
@@ -51,6 +51,7 @@
     "loc-i18next": "0.1.6",
     "papaparse": "~5.5.2",
     "rxjs": "^7.8.1",
+    "three": "^0.181.1",
```
uhon: If we use threejs more extensively in the future, it would probably make sense to use [AngularThree](https://angularthree.org/).
But maybe overkill to introduce it here.
At least would use newest version available 0.184.0

ui/src/features/controls/control-compass.element.ts
```diff
@@ -77,21 +264,47 @@ export class ControlCompass extends CoreElement {
     });
   };
 
-  readonly render = () => html`<ngm-core-icon icon="compass"></ngm-core-icon>`;
+  readonly render = () => html`
+    <div class="ngm-compass-body">
+      <canvas class="ngm-compass-canvas"></canvas>
```
uhon, line 263:
```suggestion
      <canvas class="ngm-compass-canvas" />
```

ui/src/features/controls/control-compass.element.ts
```diff
@@ -10,48 +10,235 @@ import {
...
+    this.scene3d.add(fillLight);
+
+    this.headingGroup = new Group();
+    this.pitchGroup = new Group();
+    this.rollGroup = new Group();
+    this.headingGroup.add(this.pitchGroup);
+    this.pitchGroup.add(this.rollGroup);
+    this.scene3d.add(this.headingGroup);
+
+    this.resizeObserver = new ResizeObserver(() => this.resizeRenderer());
```
uhon: Can't this be written `new ResizeObserver(this.resizeRenderer)`

ui/src/features/controls/control-compass.element.ts
```diff
@@ -10,48 +10,235 @@ import {
...
+
+        if (this.viewer !== null) {
+          this.viewer.camera.changed.removeEventListener(
+            this.handleCameraChange,
+          );
+        }
+        this.viewer = viewer;
+        this.viewer.camera.changed.addEventListener(this.handleCameraChange);
+        this.removePostRenderListener =
+          this.viewer.scene.postRender.addEventListener(
```
uhon, line 64: `handleCameraChange` is executed on every render.
Isn't it enough to only execute if camera has changed (and maybe once when the compass is initialized)?
If this needs to remain, the event listener for changed Camera values is redundant.

ui/src/features/controls/control-compass.element.ts
```diff
@@ -10,48 +10,235 @@ import {
...
+    this.pitchGroup.add(this.rollGroup);
+    this.scene3d.add(this.headingGroup);
+
+    this.resizeObserver = new ResizeObserver(() => this.resizeRenderer());
+    this.resizeObserver.observe(this);
+
+    this.resizeRenderer();
+
+    const loader = new GLTFLoader();
+    loader.load('/images/compass.gltf', (gltf: { scene: any }) => {
```
uhon: `gltf` should be of type `GLTF`.
Please add `@types/three`

ui/src/features/controls/control-compass.element.ts
```diff
@@ -77,21 +264,47 @@ export class ControlCompass extends CoreElement {
...
       align-items: center;
+      width: 48px;
+      height: 48px;
+      border-radius: 24px;
+      background: #f1f3f5;
+      box-shadow: 4px 4px 2px #00000029;
+      color: var(--ngm-interaction);
+      cursor: pointer;
 
       perspective: 600px;
```
uhon: perspective and transform-style (also in `.ngm-compas-body`) are only relevant for the old CSS transform. please remove

uhon (changes requested):
Found some issues.

Some remarks on Bundle size:

Using ThreeJS adds significant Size to the Bundle (might be relevant or not).

According to my Conversation with AI, we could use Cesium renderer (zero extra bundle cost)
Cesium is already loaded and has a full WebGL renderer. You can create a minimal off-screen Cesium.Scene on the compass <canvas> and load the GLTF via `Cesium.Model.fromGltfAsync()`. 

**Pros**: zero bundle increase, same GLTF format, full 3D, consistent rendering engine
**Cons**: Cesium's Scene has some overhead; need to manage its render loop manually; Cesium's API for off-screen scenes is less documented.

Since current approach works great and there are uncertainties with the Cesium-Approach, we can leave it IMHO as is.
Also having Three.js in the bundle might become handy in the future.

ui/src/features/controls/control-compass.element.ts
Dead0utside: `<canvas class="ngm-compass-canvas" />` triggers ESlint warning

ui/src/features/controls/control-compass.element.ts
Dead0utside: Without handleCameraChange on every render the animations become snappy and don't always fire, so it has to stay. However, it is possible to remove the event listener for changed camera values.

ui/package.json
Dead0utside: According to my conversations with an AI agent, the Three.js solution is more lightweight in terms of live performance, albeit for the price of bundle size. Although I've tried implementing this with Cesium the result was not particularly satisfactory: problems with alignment, rotation axes, texture quality etc. As for AngularThree, I don't see the appeal, since we use Lit, not Angular for the front-end. I am open to further discussion however, so let me know what you think.

uhon: Oh, sure my bad. In Lit an angular wrapper wouldn't make sense of course.
I'm fine with using Three.

uhon (approved):
Thanks for the changes. LGTM now!

PR #1956, opened 2026-05-22
feature(viewer-1953): Ergänzungen Infobox: Tooltips für lange Texte

Dead0utside:


uhon (approved):
Could not find an example to try it out. But implementation looks good to me

PR #1957, opened 2026-05-26
fix(viewer-1918): 3D Navigationsanzeige Verhältnis fix

Dead0utside:


ui/src/features/controls/control-compass.element.ts
```diff
@@ -123,9 +136,9 @@ export class ControlCompass extends CoreElement {
     this.headingGroup = new Group();
     this.pitchGroup = new Group();
     this.rollGroup = new Group();
-    this.headingGroup.add(this.pitchGroup);
-    this.pitchGroup.add(this.rollGroup);
-    this.scene3d.add(this.headingGroup);
+    this.pitchGroup.add(this.headingGroup);
```
uhon, line 139: The order of Group-Adding has changed? Is this intentional? How does it differ from the approach before?
Maybe a comment would help

uhon (approved):
It seems to work. Can't judge the Math though without more explanation.

ui/src/features/controls/control-compass.element.ts
Dead0utside: this is intentional, it fixes rotation behavior when both pitch and angle change

PR #1963, opened 2026-06-03
#1953: tooltips für alle Texte

Dead0utside:
Now the tooltips are always displayed, not only when the text is truncated

### DIFF
--- a/ui/src/features/catalog/display/details/catalog-display-legend-detail.element.ts
+++ b/ui/src/features/catalog/display/details/catalog-display-legend-detail.element.ts
@@ -125,15 +125,13 @@ export class CatalogDisplayInfoBox extends CoreElement {
             <tr>
               <td
                 class="info-key"
-                @mouseenter=${this.updateOverflowTooltip}
-                @focusin=${this.updateOverflowTooltip}
+                title=${i18next.t(`layers:info_box.information.${key}`)}
               >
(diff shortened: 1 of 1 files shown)


vej-ananas (changes requested):
Using javascript to dynamically add or remove the title is quite overkill - at the latest now that they should always be shown. You can simply set the title attribute on the `<td>` element

ui/src/features/catalog/display/details/catalog-display-legend-detail.element.ts
```diff
@@ -125,15 +125,13 @@ export class CatalogDisplayInfoBox extends CoreElement {
...
-                @focusin=${this.updateOverflowTooltip}
+                title=${i18next.t(`layers:info_box.information.${key}`)}
               >
                 ${i18next.t(`layers:info_box.information.${key}`)}
               </td>
               <td
                 class="info-value"
-                @mouseenter=${this.updateOverflowTooltip}
-                @focusin=${this.updateOverflowTooltip}
+                title=${typeof value === 'string' ? i18next.t(value) : nothing}
```
vej-ananas, line 134: Now here it gets interesting - what should the tooltext show for urls, the label or the target? This might be something that Nils should define.

vej-ananas (approved):

PR #1967, opened 2026-06-03
#1953: Infobox tooltips: only apply <a> tag if the url exists

Dead0utside:


ui/src/features/catalog/display/details/catalog-display-legend-detail.element.ts
```diff
@@ -161,6 +161,9 @@ export class CatalogDisplayInfoBox extends CoreElement {
 
   private readonly renderInformationValue = (value: InformationValue) => {
     const { url, label } = this.getInformationValueContent(value);
+    if (!url) {
+      return html`${label}`;
```
vej-ananas, line 165: i believe previously, a simple string was returned for this case so the `html` tag is not strictly necessary.

vej-ananas (approved):

PR #1968, opened 2026-06-05
#1951: adjustment - infobox item order defineable

Dead0utside:
#1951
By changing the structure of the layer configurations to contain array, it is now possible to strictly define the order of the information entries by reordering items in the layer configurations

### DIFF
--- a/layers/layers_3dtiles.json5
+++ b/layers/layers_3dtiles.json5
@@ -1031,14 +1031,20 @@
       //geocat_id: '',
       info_box: {
         source: 'custom',
-        information: {
-          contact: { key: 'contact_value', url: 'contact_url' },
-          description_sa3d_HL_profiles_01: {
-            key: 'description_value_sa3d_HL_profiles_01',
-            url: 'description_url_sa3d_HL_profiles_01',
+        information: [
+          {
+            key: 'contact',
+            value: { key: 'contact_value', url: 'contact_url' },
           },
-          dataDate: '28.04.2026',
-        },
+          {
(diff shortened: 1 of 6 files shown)


layers/layers_3dtiles.json5
```diff
@@ -448,19 +448,31 @@
...
         source: 'custom',
-        information: {
-          contact: { key: 'contact_value', url: 'contact_url' },
-          description: { key: 'description_value', url: 'description_url' },
-          download_01: {
-            key: 'download_movz_value',
-            url: 'download_movz_url_j3d_horizon',
+        information: [
+          {
+            key: 'contact',
```
vej-ananas, line 453: I would not use the same word `key` for two different targets in this context. Maybe was can call this one `labelKey`? Or maybe there are words that are even more intuitive?

ui/src/features/layer/layer-api.service.ts
```diff
@@ -110,7 +109,7 @@ export class LayerApiService extends BaseService {
           if (legendUrl !== null) {
             result.legendUrl = legendUrl;
           }
-          const information: Record<string, InformationValue> | null =
+          const information: InfoBoxCustom['information'] | null =
```
vej-ananas, line 112: nice!

vej-ananas (approved):
Looks great! Just a minor naming suggestion.

layers/layers_3dtiles.json5
Dead0utside: Good point, I'll think of a better name

PR #1973, opened 2026-06-12
#1942: Integration for any URLs

Dead0utside:
#1942 
Now the service_url is configurable per-layer.
Also adjusts the tile click -> open infobox behavior


layers/layertree.json5
```diff
@@ -26,6 +26,17 @@
             'ch.swisstopo.geologie-geotechnik-gk500-genese',
           ],
         },
+
+        // 02 - Lexic augmented layers
+        {
```
vej-ananas, line 31: I would also add the following to limit it to dev, since there is no specification that these should be added right?
```
      access: {
        env: ['local', 'dev'],
      },
```

ui/locales/layers/layers.de.json
```diff
@@ -243,6 +246,7 @@
     "industrial_minerals": "Industrieminerale",
     "j3d_ma_01": "Modellgebiet 1",
     "jura_3d_models": "swissJURA3D",
+    "lexic_augmented_layers": "Angereicherte Layer",
```
vej-ananas: Angereichert ist nicht sehr aussagekräftig, ich würde zB `Lexic Layers` hereinschreiben.

vej-ananas (approved):

layers/layertree.json5
Dead0utside: it was not explicitly mentioned

ui/locales/layers/layers.de.json
Dead0utside: Done

PR #1974, opened 2026-06-15
#1943 lexic api service

Dead0utside:
#1943


api/src/config.rs
```diff
@@ -38,6 +38,11 @@ pub struct ClientConfig {
     #[clap(long, env)]
     pub gst_url: String,
 
+    /// Optional API key for the SwissTopo WebMap Semantic Filter (Lexic) API.
+    /// Set via LEXIC_API_KEY environment variable.
+    #[clap(long, env)]
+    pub lexic_api_key: Option<String>,
```
vej-ananas: afaik it is an open API - so there will be no api key necessary.

ui/src/constants.ts
```diff
@@ -198,6 +198,14 @@ export const TITILER_BY_PAGE_HOST = {
   'viewer.swissgeol.ch': 'https://api.swissgeol.ch/titiler',
 };
 
+export const LEXIC_API_BY_PAGE_HOST = {
```
vej-ananas, line 201: we should probably set the dev api for all of these for the moment, else we will be blocked when we release.

ui/src/features/lexic/lexic-api.service.ts
```diff
@@ -0,0 +1,178 @@
...
+    this.baseUrl = LEXIC_API_BY_PAGE_HOST[this.getHost()] ?? DEFAULT_BASE_URL;
+
+    BaseService.inject$(clientConfigContext).subscribe((clientConfig) => {
+      if (clientConfig.lexicApiKey !== undefined) {
+        this.apiKey = clientConfig.lexicApiKey;
+      }
+    });
+  }
+
+  configure(config: { apiKey?: string | null; baseUrl?: string }): void {
```
vej-ananas: why is there a configure method, if the constructor already sets the values?

ui/src/features/lexic/lexic-api.model.ts
vej-ananas, line 1: How did you generate the models and service? Usually it is nice to use a library to generate these directly from the Swagger File - then it is really easy to rerun it e.g. as package.json script to quickly determince whether anything changed.

vej-ananas (changes requested):

ui/src/features/lexic/lexic-api.service.ts
Dead0utside: in case api key or the base url change during runtime, but since the api is planned to be public and the base url is static, there is indeed no need for this configure() method

api/src/config.rs
Dead0utside: removed the keys and all the related functionality

ui/src/constants.ts
Dead0utside: done

ui/src/features/lexic/lexic-api.model.ts
Dead0utside: I have added a dev dependency **Orval**. It appears to be quite capable and configurable. All the related docs are in the README

vej-ananas (approved):
This is looking great! Once the service and models are used, we might still adjust some stuff but this is a solid base line.

PR #1979, opened 2026-06-17
Remove accesstoken from URL after login

### LINKED ISSUE #1975: Security issues - Viewer
**Beschreibung**

Die in [LINK ](https://ebpgroup.atlassian.net/wiki/spaces/223443/pages/3389259777/Bearbeitung+Security-issues+aus+Pentest)beschriebenen Prio 1 Issues sollen behoben werden.

> ⚠️ **Bitte keine sicherheitsrelevanten Details in den Kommentaren / Beschreibung.** ⚠️

**Acceptance Criteria**

- [ ] Alle Prio 1 Punkte gemäss des verlinkten Dokuments sind gelöst

vej-ananas:
Resolves #1975

### DIFF
--- a/api/src/config.rs
+++ b/api/src/config.rs
@@ -38,7 +38,6 @@ pub struct ClientConfig {
     #[clap(long, env)]
     pub gst_url: String,
 
-
     #[clap(flatten)]
     pub auth: Auth,
 }
--- a/ui/src/features/session/session.service.ts
+++ b/ui/src/features/session/session.service.ts
@@ -180,6 +180,14 @@ export class SessionService extends BaseService {
     const payload = atob(accessToken.split('.')[1]);
     const claims: CognitoUser = JSON.parse(payload);
     const token = params.get('id_token') ?? '';
+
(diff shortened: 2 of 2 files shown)


uhon (approved):
This is important, so I will accept it, but this will only solves the problem in Browser-History and visible URL, but is masking the real problem rather than fixing it.

The real issue is using the Implicit Grant flow which is deprecated and considered insecure. The proper fix would be to switch to the Authorization Code flow with PKCE (Proof Key for Code Exchange), where the browser only receives a short-lived authorization code in the URL that gets exchanged server-side for tokens.

The Implicit flow was deprecated in OAuth 2.0 Security Best Current Practice because tokens in URL fragments leak through browser history, referrer headers, and other vectors. Even though the `replaceState` call cleans the URL after the fact, the tokens were still briefly exposed in the address bar and may have been captured by browser extensions, proxies, network interceptors, or analytics scripts before the cleanup fired.

We should create a ticket to implement Code Flow
Assets already uses the the Authorization Code flow, so we're save there. Maybe other repos are worth checking as well.

Btw. we should use `globalThis` instead of `window` as sonar suggested

PR #1986, opened 2026-06-22
Correct download links and translations

nOester:
@Dead0utside 

- Corrections of download links
- Corretions of translation

#1966


Dead0utside (approved):
lgtm

PR #1993, opened 2026-06-24
#1992: chore - deactivate endpoints

Dead0utside:
#1992


api/src/lib.rs
```diff
@@ -78,22 +78,22 @@ pub async fn app(pool: PgPool) -> Router {
...
-            get(handlers::get_project)
-                .put(handlers::update_project)
-                .delete(handlers::delete_project),
-        )
-        .route(
-            "/api/projects/:id/geometries",
-            put(handlers::update_project_geometries),
-        )
-        .route("/api/projects/upload_asset", post(handlers::upload_asset))
+//         .route(
```
vej-ananas: Delete them please instead of commenting. They will be available in the git history.

vej-ananas (changes requested):

vej-ananas (approved):

PR #2007, opened 2026-07-01
#1940 fix: display lexic layers in correct styling

Dead0utside:
#1940 fix - use the correct styling for lexic layers


uhon (approved):
LGTM

Issue #2009, opened 2026-07-02
lexic | Implement lexic vocabulary service


Dead0utside:
**Beschreibung**

The service should take care of fetching the vocabulary

**Abgrenzung**

**Acceptance Criteria**

PR #2010, opened 2026-07-02
Feature/viewer 2009: lexic vocabulary service

Dead0utside:
#2009


ui/src/features/lexic/lexic-url.test.ts
```diff
@@ -0,0 +1,256 @@
...
+  getLexicHref,
+  isLexicTermUrl,
+  parseLexicTermUrl,
+  toLexicLanguage,
+} from './lexic-url';
+
+describe('parseLexicTermUrl', () => {
+  it('parses a Chronostratigraphy URL', () => {
+    const result = parseLexicTermUrl(
+      'https://dev-lexic.swissgeol.ch/Chronostratigraphy/LateBurdigalian',
```
uhon: URLs repeat quite often.
If they are the same, i would store them in a constant per test.
Would make it much more readable. Otherwise we have to check wheter the whole URL (or mnost of it) are actually equal.

I would even do it when we want to test for query-params removal (easier on the eye):
````
it('strips query parameters from the normalized URL', () => {
  const expectedURL = 'https://dev-lexic.swissgeol.ch/Lithology/SandstoneGlauconite`,    
  const result = parseLexicTermUrl(expectedURL + "?lang=de");
    expect(result).toEqual({
      vocabularyId: 'lithology',
      normalizedTermUrl:
        expectedURL',
    });
  });
````

File: ui/src/features/layer/info/layer-info.model.ts [hunk 2]
```diff
@@ -47,12 +47,29 @@ export interface LayerInfo {
 
 export interface LayerInfoAttribute {
   key: string | TranslationKey;
-  value: LayerInfoValue | LayerInfoUrl;
+  value: LayerInfoValue | LayerInfoUrl | LayerInfoLexicTerm;
```
uhon: This leads to a compile error!
Please check all places where value is used

ui/src/features/lexic/lexic-url.ts
```diff
@@ -0,0 +1,98 @@
...
+const LEXIC_HOSTS = new Set([
+  'dev-lexic.swissgeol.ch',
+  'int-lexic.swissgeol.ch',
+  'lexic.swissgeol.ch',
+]);
+
+/** Example
+ * 'https://dev-lexic.swissgeol.ch/Lithostratigraphy/StGallenFormation?lang=de'
+ * vocabularyId = 'lithostratigraphy'
+ * normalizedTermUrl = 'https://dev-lexic.swissgeol.ch/Lithostratigraphy/StGallenFormation?lang=de'
```
uhon: Comment says: "language query parameter removed", but it isn't. (code ok, but this commnet is wrong

ui/src/features/lexic/lexic-url.ts
```diff
@@ -0,0 +1,98 @@
...
+ * (language query parameter removed)
+ */
+export interface ParsedLexicTermUrl {
+  vocabularyId: string;
```
uhon: why `string`, this should be: `LexicVocabularyId`

ui/src/features/lexic/lexic-vocabulary.service.test.ts
```diff
@@ -0,0 +1,360 @@
...
+import { LexicVocabularyService } from './lexic-vocabulary.service';
+import { LexicApiService } from './lexic-api.service';
+import {
+  getLexicHref,
+  isLexicTermUrl,
+  parseLexicTermUrl,
+} from 'src/features/lexic/lexic-url';
+
+describe('parseLexicTermUrl', () => {
+  it('parses a Chronostratigraphy URL', () => {
```
uhon: Don't we cover these URL tests already in `lexic-url.test.ts`?

uhon (changes requested):
Found a few things.

Caching mechanism is solid, but could be optimized by using localStorage with a reasonable TTL. otherwise we issue 4 dictionary requests on every page reload for things that rarely change

uhon (approved):
Looks good now.

Nice solution with `makeTerm` and `mockTerms`

PR #2013, opened 2026-07-10
allow to fetch single tile for lexic filter

uhon:
This PR makes Lexic Filter request a single tile instead of default multi-tile behavior.

It's configurable via Query-params. `?lexicSingleTile=false` disables this behavior and will fall back to fetching multiple tiles (as before)

**Attention:** Using a single tile does not make the browser refresh any overlay content, as with multiple tiles. But with such a slow WMS endpoint, reloading tiles wouldn't make sense anyways


uhon:
We agreed on not (yet) merging this functionality. Keep it as a reference

vej-ananas (approved):

uhon:
Now mergeable. We agreed on switching default behavior (defaults to tiled image request).

Issue #1954, opened 2026-05-22
lexic | Advanced search - Filter component


wendlans:
**Beschreibung**

The Viewer integrates semantic filter capabilities from the Lexic API, allowing users to filter map layers by different mechanics:
* #1944
* #1945
* #1946
* #1947
* #1948

To access these filters there is a new button on the left side:
<img width="248" height="265" alt="Image" src="https://github.com/user-attachments/assets/d11fa35e-b15f-47b8-b5fd-0d2480aa8a05" />

Which then opens a new panel: Lexic filter
<img width="361" height="781" alt="Image" src="https://github.com/user-attachments/assets/b6798552-ff0e-4a57-9fea-a145b9812421" />

The goal of this ticket is to create a new button and the corresponding new Lexic filter panel. The Lexic filter panel is used as a container for:
* **choice of layer (Datensatz)**
By default the first layer will be pre-selected. In the future it should be possible to set this value from the outside. The reason is that we want to implement the filter symbol on the Lexic layer in the "Anzeige" as well - which opens the Lexic filter panel with the selected layer. But for now just pre-select the first layer.
* **the Lexic filters**
  * result filter component #2002
  * all supported filters for the chosen layer (Datensatz)

The component should already include the dropdown to chose the layer (Datensatz). All available layers can fetched from the Lexic API: https://ebpgroup.atlassian.net/wiki/spaces/223443/pages/3248128005/Einbindung+Lexic#0.-Discover-supported-layers
This will most likely happen multiple times so a local cache could be useful here.

Figma: https://www.figma.com/design/cEiOoOazAQZqpRY92ZhBeO/swissgeol?node-id=20103-253075&t=Ku3WEaS8h5kx9tQb-0
Further details: https://ebpgroup.atlassian.net/wiki/spaces/223443/pages/3248128005/Einbindung+Lexic#Viewer-Advanced-search-(Lexic-API)

**Abgrenzung**

* The individual filters are created in their respective tickets and are not part of this ticket.
* The result layer is handled within its own ticket.

**Acceptance Criteria**

- [ ] The Lexic button on the left side is visible
- [ ] The Lexic filter component is added and opens upon selecting the above button
- [ ] The supported layers (Datensatz) can be chosen from

nOester:
@wendlans 
Parametrisierte Übergabe an die Filterbox

uhon:
@wendlans
[...]

PR #2014, opened 2026-07-10
Feature/viewer 1954 improvements to filter component

uhon:
This fixes the latest reported bugs from #1954

Summary:

- Only shows supported filters (currently lithology, tectonic-units and lithostratigraphy). 
@wendlans I did not do any extra work for tectonic-units and lithostratigraphy just had to add them to the list of supported filters, since their behavior is equal to lithology and our generic components just handle them correctly.
- Layers without any supported filter are hidden from datasets
- Fixed English translation from "dataset" to "Dataset".
- Persist selected dataset and filters across language changes 
- Closing panel now clears all filters and resets the selected dataset. Changing the dataset also clears filters.
- Retranslates existing filter labels on language change
- Excludes already-applied filters from filter dialog


Dead0utside (approved):
lgtm

Issue #2001, opened 2026-06-29
lexic | Direct link Viewer to Lexic - Translation


wendlans:
**Beschreibung**

Für den Direktlink vom Viewer zu Lexic müssen alle dazugehörigen Komponenten übersetzt werden:

<img width="425" height="470" alt="Image" src="https://github.com/user-attachments/assets/b1c883c3-88df-4865-8080-44e8b0f0cf2a" />

1. **Namen der Zeilenheaders:** Lokale Übersetzung analog zu anderen Layern.
2. **Link-Label:** Anstelle von "Link" soll der übersetzte Name des Vokabular-Terms angezeigt werden. Siehe https://ebpgroup.atlassian.net/wiki/spaces/223443/pages/3248128005/Einbindung+Lexic#Viewer-Direct-link-from-Viewer-to-Lexic-(URL)
3. **Name des Layers:** Lokale Übersetzung analog zu anderen Layern.

**Acceptance Criteria**

- [ ] Das Vokabular muss lokal gecached werden
- [ ] Alle Einträge im Info-Panel sind übersetzt:
  - [ ] Namen der Zeilenheaders
  - [ ] Link-Label + URL
  - [ ] Name des Layers

Dead0utside:
Die Übersetzungen für den Infobox Bezeichnungen sind laut @wendlans implementiert. Die sind lokal gespeichert und können jederzeit verändert werden, falls es benötigt wird

<img width="424" height="332" alt="Image" src="https://github.com/user-attachments/assets/4766229e-c9f2-47e8-a817-f92cd2906810" />

wendlans:
@Dead0utside I've asked the Lexic developers about the current translation. One question for you though:
What about the translation of the *3*?
<img width="425" height="470" alt="Image" src="https://github.com/user-attachments/assets/b1c883c3-88df-4865-8080-44e8b0f0cf2a" />

Dead0utside:
@wendlans it is also stored in our local translation files. The data was taken from Lexic API

wendlans:
@Dead0utside Perfect. I've got answers about the translation:
> Those german translations appear to be correct.
>
> They are exhaustive. In GC_BEDROCK all possible vocabularies are present.

wendlans:
@Dead0utside After switching the language the labels won't get updated (however, the values - after some time - will):

[...]

PR #2015, opened 2026-07-10
Fix/ viewer-2001 - labels not changing on language change

Dead0utside:
#2001 
fix: on language change refetch/reload translations

### DIFF
--- a/ui/src/features/layer/info/layer-info.service.ts
+++ b/ui/src/features/layer/info/layer-info.service.ts
@@ -13,6 +13,7 @@ import {
   Observable,
   take,
 } from 'rxjs';
+import i18next from 'i18next';
 import { BaseService } from 'src/services/base.service';
 import { Layer, LayerType } from 'src/features/layer';
 import {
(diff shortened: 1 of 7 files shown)


ui/src/features/layer/info/pickers/layer-info-picker-for-wmts.ts
```diff
@@ -212,25 +214,39 @@ export class LayerInfoPickerForWmts implements LayerInfoPicker {
...
+    this.identifyResult = data.identifyResult ?? null;
+    this.service = data.service ?? null;
     this.dataSource.entities.add(this.entity);
   }
 
+  async refreshForLanguage(lang: string): Promise<void> {
+    if (this.identifyResult == null || this.service == null) {
+      return;
+    }
+    const html = await this.service.fetchHtmlPopup(this.identifyResult, lang);
```
uhon: What if this fails? don't we have to catch this?
also changing languages rapidly might lead to race-conditions (multiple fetchHtmlPopup calls. I think we can ignore this, but we need to catch potential errors

ui/src/features/layer/info/layer-info.service.ts
```diff
@@ -74,12 +75,32 @@ export class LayerInfoService extends BaseService {
           ScreenSpaceEventType.LEFT_CLICK,
         );
       });
+
+    i18next.on('languageChanged', this.handleLanguageChanged);
```
uhon, line 79: Don't we have to remove this event listener when destroying the component?

uhon (changes requested):

ui/src/features/layer/info/layer-info.service.ts
uhon: It's a singleton anyways. we keep it as is

ui/src/features/layer/info/pickers/layer-info-picker-for-wmts.ts
uhon: fixed

uhon (approved):

PR #2031, opened 2026-08-03
Fixing tile request errors of missmatching bounding boxes.

uhon:
Fixing tile request errors of missmatching bounding boxes.
The server responded with "Row X is out of range" per zoom level, because those tile rows don't exist in its grid

Reason is, that the tilebox was requesting Switzerlands Bounding box (as other layers) do,
The fix: read the layer's actual bounding box from the capabilities

GC_BEDROCK uses a smaller bounding box, hence the unavailable tiles. The actual bounding box used is passed in the WMTS capabilities document, we now parse it and pass it on to Cesium, so it only requests tiles that actually exist.d

This also solves the layer-clicking behavior. Info-Box appears directly now.

Also fixes:
removed double URL encoding for layer names - fixes loading certain layers via URL

### DIFF
--- a/ui/locales/layout/layout.de.json
+++ b/ui/locales/layout/layout.de.json
@@ -17,7 +17,7 @@
       "or": "OR",
       "dialogTitle": "Filter – {{title}}",
       "searchPlaceholder": "{{title}} suchen ...",
-      "includeNarrowers": "Untergeordnete einschliessen",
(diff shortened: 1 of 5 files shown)


Dead0utside (approved):
looks good. I'd make adjustments, suggested by sonar. Doesn't seem like they are difficult to implement. But we can also merge as it is, I see no problems

PR #2041, opened 2026-08-07
Feature/viewer 2024 cesium update and new camera controller

### LINKED ISSUE #2024: Neuen CameraController integrieren
**Beschreibung**

Der bestehende Default-Camera-Controller soll mit dem neuen CameraController von CesiumJS soll ersetzt werden.

- Weitere Informationen: https://github.com/swisstopo/swissgeol-viewer-suite/issues/1854
- Sandcastle: https://ci-builds.cesium.com/cesium/screenspace-camera-controllers/Apps/Sandcastle2/index.html?id=camera-controllers
- Beschreibung Implenentierung: [screenspace-camera-controllers ](https://github.com/CesiumGS/cesium/pull/13604)
- Ref docs:
  - [Viewer.addController](https://ci-builds.cesium.com/cesium/screenspace-camera-controllers/Build/Documentation/Viewer.html?classFilter=viewer#addController)/ Viewer.removeController
  - [Scene.controllerHost](https://ci-builds.cesium.com/cesium/screenspace-camera-controllers/Build/Documentation/Scene.html#controllerHost)
  - [Controller interface](https://ci-builds.cesium.com/cesium/screenspace-camera-controllers/Build/Documentation/Controller.html), with links to each controller
  - [ScreenSpaceInputBindings](https://ci-builds.cesium.com/cesium/screenspace-camera-controllers/Build/Documentation/ScreenSpaceInputBindings.html?classFilter=Input)

**Hinweis:** Der neue CameraController ist noch nicht im main-Branch von CesiumJS enthalten. Geplant ist, dass er in den August-release kommt (03.08.2026)

Vorgänger-Ticket: #1905 

**Abgrenzung**

**Acceptance Criteria**
- [ ] Die Navigation der Camera läuft über den neuen CameraController

uhon:
### Summary

Updates CesiumJS from 1.135.0 to 1.144.0 and migrates camera input handling from the monolithic `ScreenSpaceCameraController` to the new modular controller API introduced in [CesiumGS/cesium#13604](https://github.com/CesiumGS/cesium/pull/13604) (`viewer.addController()`/`removeController()`). Also ports keyboard navigation to a proper service, and fixes two long-standing bugs uncovered along the way (nadir heading-flip, map-freeze on failed picks).

Closes #2024.

### Changes (see individual commits)

1. **Update CesiumJS to 1.144.0** — dependency bump only, enables the new controller API.
2. **Implement new modular camera controllers** (`CameraControllerService`)
   - `ScreenSpaceMapCameraController` for pan, `ScreenSpaceTiltOrbitCameraController` for tilt/orbit (middle-drag or Ctrl+left-drag), `ScreenSpaceZoomCameraController` for zoom (scroll + right-drag).
   - Old `ScreenSpaceCameraController` inputs are disabled entirely; `CameraControllerService` exposes `addController()`/`removeController()` and an `enableInputs` getter/setter so features can add/remove or temporarily suspend camera input (used by draw tools, slicer arrows, camera locks, tracking, 2D mode).
   - `Control2dController`, `ngm-cam-configuration`, `ngm-nav-tools`, `CesiumDraw`, `SlicerArrows` refactored off the old controller's ad-hoc flags (`enableRotate`, `tiltEventTypes` backup/restore, `enableCollisionDetection`, etc.).
3. **Implement keyboard navigation service** (`KeyboardNavigationService`)
   - Ports the old `KeyboardNavigation.js` (WASD/QE + look keys) to a typed, `@lit/context`-registered service.
   - Vertical movement (`Q`/`E`) uses the local ENU (East-North-Up) frame, so "up" always means away from the globe regardless of camera tilt.
   - Movement speed is adjustable via `Shift`+scroll wheel, and additionally **auto-scales with camera altitude**: lateral movement (A/D) scales roughly with `sqrt(height)`, forward/back/vertical movement (W/S/Q/E) scales linearly with height (both floored at 1× near the surface). This keeps navigation usable at both ground level and continental-scale altitudes without needing to constantly readjust speed.
4. **Fix world picking from the underside** (`pickWorldPositionWithDepthBuffer`)
   - The controllers' default `pickWorldPosition` only picks against the ellipsoid/plane, which breaks drag targeting when the camera is underground or below terrain overhangs.
   - Added a custom pick function that tries the depth buffer (`scene.pickPosition`, actual rendered geometry/terrain) first and falls back to the ellipsoid pick; wired into tilt, zoom and pan controllers.
5. **Patch `ScreenSpaceTiltOrbitCameraController` — nadir gimbal-lock fix**
   - Root cause (verified directly in `@cesium/engine`'s `Camera.js`): both `tilt()` and `orbit()` end by calling `camera.lookAtWorldPosition()`, which reconstructs the camera's `right`/`up` vectors via `Cartesian3.cross(direction, worldUp)`. At exact nadir/zenith (pitch = ±90°) this cross product degenerates to ~0, so Cesium falls back to reusing the *previous frame's* `right` vector — which flips sign unpredictably due to floating-point noise, causing a visible 180° heading flip while tilting/orbiting straight down.
   - `patchTiltNadirGimbalLock()` monkey-patches both methods on our controller instance, keeping the exact original damping/velocity/position math but replacing the final `lookAtWorldPosition()` call with a direct rotation of `direction`/`up`/`right` by the same rotation matrix (the same technique Cesium's own `Camera.prototype.rotate()` uses). This never reconstructs orientation from a fixed world-space reference, so it stays continuous through the pole and can't flip.
   - Reproduced and confirmed fixed via a stripped-down Sandcastle (linked in the issue).
6. **Fix long-standing map-freeze bug on failed scene picks**
   - `scene.pickPosition()` can throw while tiles/primitives aren't fully loaded yet, or after the scene/camera was destroyed mid-pick. `PickService` already had a fallback for this, but only matched Firefox's error wording (`"Can't access property _target, v3 is undefined"`) — Chrome/V8 throws a differently worded message (`"Cannot read properties of undefined (reading '_target')"`), which propagated as an **uncaught exception**, aborting the in-progress `pointermove`/drag handler and freezing camera interaction until reload.
   - Added the Chrome/V8 message variant, factored the check into an exported `isKnownScenePickingError()` helper, and reused it in `pickWorldPositionWithDepthBuffer` (which previously had no error handling on its `scene.pickPosition()` call at all).
   - Also changed the scene's `renderError` handler to resume rendering asynchronously (`setTimeout`) instead of calling `requestRender()` synchronously, to avoid potential infinite error loops while destroyed resources are cleaned up.

### What was removed / thrown out

- Old `src/KeyboardNavigation.js` (replaced by `KeyboardNavigationService`).
- Ad-hoc camera-controller manipulation throughout the codebase: `tiltEventTypes`/`lookEventTypes` backup-and-restore in `Control2dController`, direct `scene.screenSpaceCameraController.enable*` flag toggling in `ngm-cam-configuration`, `ngm-nav-tools`, `CesiumDraw`, `SlicerArrows` — all replaced by `CameraControllerService.enableInputs` / `addController()` / `removeController()`.


uhon:
@volodymyr-h 

If you don't find issues, feel free to merge it after accepting on Monday, so that Nils can test it on DEV.
If any issue found, I'll fix it on Tuesday when I'm back in the office.

Thank you

ui/src/features/controls/keyboard-navigation.service.ts
```diff
@@ -0,0 +1,283 @@
...
+      hasRotation = true;
+    }
+    if (hasRotation) {
+      camera.setView({ orientation: { heading, pitch } });
+    }
+
+    this.scene.requestRender();
+  }
+
+  destroy(): void {
```
Dead0utside, line 226: is it not necessary to also remove the event listeners on canvas?
Something like:
```
if (this.canvas) {
  this.canvas.removeEventListener('wheel', this.wheelHandler, { capture: true } as EventListenerOptions);
  this.canvas.removeEventListener('click', this.onCanvasClick, { capture: false } as EventListenerOptions);
  this.canvas = undefined;
}
```

what do you think?

Dead0utside (commented):
Looks good with one small point

ui/src/features/controls/keyboard-navigation.service.ts
uhon: good catch! Fixed it

Dead0utside (approved):

PR #2047, opened 2026-08-14
feat(ui): seismic 3D image-slice planes

uhon:
Shows seismic layers as three interactive image-slice planes

Adds a "slice" mode for 3D-tiles seismic layers: instead of the full tileset, the layer renders as three orthogonal planes (inline, crossline, depth), each showing one image slice selected via HUD sliders. Slices are preloaded/cached in the background so switching between them is instant.

UI:
- New catalog detail panel with per-axis sliders (catalog-display-slice-detail.element.ts) to pick a slice per plane; multi-slice mode is present but disabled with a "coming soon" label.
- Sliders are keyboard-controllable (arrow/Home/End/PageUp/PageDown).
- Preload progress banner: subtle/low-key styling, hidden if slices were already cached, shown while loading, lingers briefly after completion.
- `sliceSelection` added to the Tiles3dLayer model, wired through catalog, Ion, and layer API mapping.
- Fixed a "zoom to object" button regression where it could get stuck disabled for slice layers.

Slice loading & caching (features/layer/slice/):
- Resolve and cache per-slice tileset JSON/metadata, tolerating the OGC gateway's redirect-to-S3 behaviour and occasional outages (falls back to a known public S3 URL for a real, deployed seismic layer).
- Double-buffered tileset swapping per axis with generation-counter based cancellation and in-flight de-duplication, so rapid slider changes never race or reuse a tileset mid-destroy.
- Background HTTP-cache warming and GPU tileset pre-building around the current selection, gated to only run while the HUD panel is open (pausing preserves already-fetched progress; only full layer teardown discards it), so reopening the panel doesn't needlessly reload slices.
- Disabled the tileset's environment-map bake per slice tileset to avoid transient WebGL errors from it being interrupted by fast tileset destruction while slices are swapped.
- LRU pruning of built tilesets to bound memory/GPU usage.

Picking & reliability fixes:
- pick.service.ts: fall back to an ellipsoid intersection when the primary terrain ray-cast misses (e.g. camera underground looking away from terrain) instead of aborting the pick — lets underground content (slice planes) be clicked reliably instead of only when the ray happens to hit terrain.
- layer-info-picker-for-wmts.ts: skip the slow, network-bound geo.admin identify request while the camera is underground.
- layer-info-picker-for-wmts.service.ts: 6s timeout on all geo.admin fetches, so a hanging/slow request can no longer stall the info box indefinitely.
- layer.controller.ts: the per-layer update queue now propagates/rejects on error instead of silently stalling.
- layer.service.ts: fixed a cold-start race where a fast session init could call loadLayers() before the layer API client was injected.

### DIFF
--- a/ui/src/features/catalog/display/catalog-display-list-item.element.ts
+++ b/ui/src/features/catalog/display/catalog-display-list-item.element.ts
@@ -12,6 +12,7 @@ import {
   AnyLayer,
   getLayerLabel,
   isBackgroundLayer,
+  isDefaultSliceSelection,
   Layer,
   LayerType,
   Tiles3dLayerController,
@@ -69,20 +70,32 @@ export class CatalogDisplayListItem extends CoreElement {
       return;
     }
 
-    if (controller instanceof Tiles3dLayerController) {
-      this.canZoom = !!controller.tileset?.boundingSphere;
-      if (!this.canZoom) {
-        const checkInterval = setInterval(() => {
-          if (controller.tileset?.boundingSphere) {
-            this.canZoom = true;
-            this.requestUpdate();
-            clearInterval(checkInterval);
-          }
-        }, 100);
-      }
-    } else {
+    if (!(controller instanceof Tiles3dLayerController)) {
       this.canZoom = true;
+      return;
     }
+
+    this.canZoom = !!controller.tileset?.boundingSphere;
+    if (this.canZoom) {
+      return;
+    }
+
+    // Tileset and slice metadata finish loading asynchronously after activation.
+    const startedAt = Date.now();
+    const checkInterval = setInterval(() => {
+      const isReady = !!controller.tileset?.boundingSphere;
+      if (isReady) {
+        this.canZoom = true;
+      }
+      if (
+        controller.supportsSliceSelection ||
+        (isReady && Date.now() - startedAt > 2_000) ||
+        Date.now() - startedAt > 15_000
+      ) {
+        this.requestUpdate();
+        clearInterval(checkInterval);
+      }
+    }, 100);
   }
 
   updated() {
@@ -136,8 +149,10 @@ export class CatalogDisplayListItem extends CoreElement {
       ...options,
       onClose: () => {
         this.windows[name] = null;
+        this.requestUpdate();
       },
     });
+    this.requestUpdate();
   }
 
   private readonly openLegend = (): void =>
@@ -183,6 +198,47 @@ export class CatalogDisplayListItem extends CoreElement {
       `,
     });
 
+  private readonly openSlice = (): void =>
+    this.openWindow('slice', {
+      title: () =>
+        i18next.t('catalog:slice_window.title', {
+          layer: getLayerLabel(this.layer),
+        }),
+      body: () => html`
+        <ngm-catalog-display-slice-detail
+          .layerId=${this.layer.id}
+        ></ngm-catalog-display-slice-detail>
+      `,
+    });
+
+  private get tiles3dController(): Tiles3dLayerController | null {
+    if (this.layer.type !== LayerType.Tiles3d) {
+      return null;
+    }
+    const controller = this.layerService.controller(this.layer.id);
+    return controller instanceof Tiles3dLayerController ? controller : null;
+  }
+
+  private get supportsSliceSelection(): boolean {
+    return this.tiles3dController?.supportsSliceSelection === true;
+  }
+
+  private get isSliceFilterActive(): boolean {
+    if (!this.supportsSliceSelection) {
+      return false;
+    }
+    if (this.windows.slice !== null) {
+      return true;
+    }
+    const selection =
+      this.layer.type === LayerType.Tiles3d ? this.layer.sliceSelection : null;
+    const defaults = this.tiles3dController?.getDefaultSliceSelection() ?? null;
+    if (selection === null || defaults === null) {
+      return false;
+    }
+    return !isDefaultSliceSelection(selection, defaults);
+  }
+
   private readonly handleOpacityChangeEvent = throttle(
     (event: SliderChangeEvent): void => {
       this.layerService.update(this.layerId, { opacity: event.detail.value });
@@ -212,7 +268,6 @@ export class CatalogDisplayListItem extends CoreElement {
         </ngm-core-button>
 
         <span class="title">${title}</span>
-
         <div class="suffix">
           ${when(
             isBackgroundLayer(this.layer),
@@ -241,6 +296,23 @@ export class CatalogDisplayListItem extends CoreElement {
             ${Math.round(this.layer.opacity * 100)}%
           </ngm-core-button>
           ${tooltip(i18next.t('catalog:display.opacity'))}
+          ${when(
+            this.supportsSliceSelection,
+            () => html`
+              <ngm-core-button
+                transparent
+                variant="tertiary"
+                shape="icon"
+                class="slice-filter"
+                ?active="${this.isSliceFilterActive}"
+                data-cy="slice-filter"
+                @click="${this.openSlice}"
+              >
+                <ngm-core-icon icon="filter"></ngm-core-icon>
+              </ngm-core-button>
+              ${tooltip(i18next.t('catalog:display.slice'))}
+            `,
+          )}
           ${when(!isBackgroundLayer(this.layer), this.renderActions)}
         </div>
       </div>
@@ -317,6 +389,15 @@ export class CatalogDisplayListItem extends CoreElement {
           >
             <ngm-core-icon icon="filter"></ngm-core-icon>
             ${i18next.t('catalog:display.filter')}
+          </ngm-core-dropdown-item>
+        `,
+      )}
+      ${when(
+        this.supportsSliceSelection,
+        () => html`
+          <ngm-core-dropdown-item role="button" @click="${this.openSlice}">
+            <ngm-core-icon icon="filter"></ngm-core-icon>
+            ${i18next.t('catalog:display.slice')}
           </ngm-core-dropdown-item>
         `,
       )}
@@ -536,7 +617,7 @@ export class CatalogDisplayListItem extends CoreElement {
   `;
 }
 
-type WindowName = 'legend' | 'times' | 'voxelFilter' | 'tiffFilter';
+type WindowName = 'legend' | 'times' | 'voxelFilter' | 'tiffFilter' | 'slice';
 
 type WindowMapping = Record<WindowName, CoreWindow | null>;
 
@@ -560,6 +641,7 @@ const getWindowsOfLayer = (layerId: Id<AnyLayer>): WindowMapping => {
     tiffFilter: null,
     times: null,
     voxelFilter: null,
+    slice: null,
   };
   windowMappingsByLayerId.set(layerId, newMapping);
   return newMapping;
--- a/ui/src/features/catalog/display/catalog-display-list-item.element.ts
+++ b/ui/src/features/catalog/display/catalog-display-list-item.element.ts
@@ -48,6 +48,7 @@ export class CatalogDisplayListItem extends CoreElement {
   accessor canZoom = true;
 
   private windows!: WindowMapping;
+  private canZoomPollIntervalId: ReturnType<typeof setInterval> | null = null;
 
   connectedCallback() {
     super.connectedCallback();
@@ -63,7 +64,21 @@ export class CatalogDisplayListItem extends CoreElement {
     );
   }
 
+  disconnectedCallback() {
+    super.disconnectedCallback();
+    this.stopCanZoomPolling();
+  }
+
+  private stopCanZoomPolling(): void {
+    if (this.canZoomPollIntervalId !== null) {
+      clearInterval(this.canZoomPollIntervalId);
+      this.canZoomPollIntervalId = null;
+    }
+  }
+
   private updateCanZoom(): void {
+    this.stopCanZoomPolling();
+
     const controller = this.layerService.controller(this.layerId);
     if (!controller) {
       this.canZoom = false;
@@ -80,19 +95,25 @@ export class CatalogDisplayListItem extends CoreElement {
       return;
     }
 
-    // Tileset and slice metadata finish loading asynchronously after activation.
-    const startedAt = Date.now();
-    const checkInterval = setInterval(() => {
+    // Tileset and slice metadata finish loading asynchronously after
+    // activation. Poll until readiness is actually observed rather than
+    // giving up after a fixed timeout — otherwise a tileset that becomes
+    // ready after the timeout would leave the zoom button permanently
+    // disabled. Polling is stopped as soon as it succeeds, the layer/
+    // controller changes (see `updateCanZoom` call above), or the element
+    // disconnects (see `disconnectedCallback`).
+    this.canZoomPollIntervalId = setInterval(() => {
+      if (this.layerService.controller(this.layerId) !== controller) {
+        // Controller was swapped out (e.g. layer deactivated/reactivated) —
+        // the subscription above will have already called `updateCanZoom()`
+        // for the new controller, so just stop this stale poll.
+        this.stopCanZoomPolling();
+        return;
+      }
       const isReady = !!controller.tileset?.boundingSphere;
       if (isReady) {
         this.canZoom = true;
-      }
-      if (
-        (isReady && Date.now() - startedAt > 2_000) ||
-        Date.now() - startedAt > 15_000
-      ) {
-        this.requestUpdate();
-        clearInterval(checkInterval);
+        this.stopCanZoomPolling();
       }
     }, 100);
   }
--- a/ui/src/features/catalog/display/catalog-display-list-item.element.ts
+++ b/ui/src/features/catalog/display/catalog-display-list-item.element.ts
@@ -88,7 +88,6 @@ export class CatalogDisplayListItem extends CoreElement {
         this.canZoom = true;
       }
       if (
-        controller.supportsSliceSelection ||
         (isReady && Date.now() - startedAt > 2_000) ||
         Date.now() - startedAt > 15_000
       ) {
--- a/ui/src/features/catalog/display/details/catalog-display-slice-detail.element.ts
+++ b/ui/src/features/catalog/display/details/catalog-display-slice-detail.element.ts
@@ -389,7 +389,7 @@ export class CatalogDisplaySliceDetail extends CoreElement {
           return null;
         }
         const min = numbers[0];
-        const max = numbers[numbers.length - 1];
+        const max = numbers.at(-1)!;
         const value = this.getSingleValue(axis);
         return html`
           <div class="axis">
--- a/ui/src/features/catalog/display/details/catalog-display-slice-detail.element.ts
+++ b/ui/src/features/catalog/display/details/catalog-display-slice-detail.element.ts
@@ -126,12 +126,22 @@ export class CatalogDisplaySliceDetail extends CoreElement {
   }
 
   disconnectedCallback(): void {
+    clearTimeout(this.preloadBannerHideTimer);
+    // Commit any pending slider value *before* tearing down HUD preloading
+    // state below. `commitDraftSingle()` may call `layerService.update()`,
+    // which emits synchronously (`BehaviorSubject.next()`), re-entrantly
+    // re-running the `layer$` subscription from `connectedCallback()` — its
+    // teardown (via `super.disconnectedCallback()`) has not run yet at this
+    // point. Committing first means that reentrant run still sees
+    // `hudActiveController`/`preloadSubscriptionAxis` pointing at the same
+    // controller, so it is a no-op instead of turning HUD preloading back on.
+    this.cancelScheduledCommit();
+    this.commitDraftSingle();
+
     this.hudActiveController?.setHudActive(false);
     this.hudActiveController = null;
     this.preloadSubscriptionAxis = null;
-    clearTimeout(this.preloadBannerHideTimer);
-    this.cancelScheduledCommit();
-    this.commitDraftSingle();
+
     super.disconnectedCallback();
   }
 
--- a/ui/src/features/catalog/display/details/catalog-display-slice-detail.element.ts
+++ b/ui/src/features/catalog/display/details/catalog-display-slice-detail.element.ts
@@ -23,6 +23,40 @@ import { when } from 'lit/directives/when.js';
 /** How long the preload banner stays visible after finishing, before fading out. */
 const PRELOAD_BANNER_LINGER_MS = 3_000;
 
+/**
+ * Slice numbers are not guaranteed to be contiguous (e.g. `[10, 20, 30]`), so
+ * a plain min/max range with a step of 1 would let the slider/stepper select
+ * values that do not actually exist. Snap to the closest value that is
+ * actually present instead. Assumes `numbers` is sorted ascending and
+ * non-empty.
+ */
+const nearestSliceNumber = (
+  numbers: readonly number[],
+  value: number,
+): number =>
+  numbers.reduce((closest, candidate) =>
+    Math.abs(candidate - value) < Math.abs(closest - value)
+      ? candidate
+      : closest,
+  );
+
+/**
+ * Index of the closest entry in `numbers` to `value`, offset by `step`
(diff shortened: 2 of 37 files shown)


ui/src/features/layer/slice/tileset-slice-metadata.ts
```diff
@@ -0,0 +1,200 @@
...
+      }
+    }
+
+    for (const child of tile.children ?? []) {
+      visit(child);
+    }
+  };
+  visit(json.root);
+
+  const hasDirectionalMeta = (['u', 'v', 'w'] as OgcSliceDirection[]).some(
```
Dead0utside: Is it guaranteed that `sliceDirection/sliceNumber` metadata is either present for all slice tiles or absent for all of them? As implemented, the presence of directional metadata on a single tile disables the URI fallback globally, so URI-only slices in a partially annotated tileset would be silently dropped.

ui/src/features/layer/slice/tileset-slice-metadata.ts
```diff
@@ -0,0 +1,200 @@
...
+const readSliceIdentity = (
+  tile: TilesetTileNode,
+): { direction: OgcSliceDirection; number: number } | null => {
+  const props = tile.metadata?.properties;
+  if (props === undefined) {
+    return null;
+  }
+  const direction = props['sliceDirection'];
+  const number = props['sliceNumber'];
+  if (!isOgcSliceDirection(direction) || typeof number !== 'number') {
```
Dead0utside: Should we validate sliceNumber with Number.isFinite() / Number.isInteger() here? With the current typeof number === 'number' check, values like NaN and Infinity would still be accepted.

ui/src/features/layer/slice/tileset-slice-metadata.test.ts
```diff
@@ -0,0 +1,270 @@
+import { describe, expect, it } from 'vitest';
```
Dead0utside: if the case with mixed types of metadata (see one of the previous comments) is realistic, it must be covered in tests

ui/src/features/layer/slice/ogc-tileset-resolve.ts
```diff
@@ -0,0 +1,213 @@
...
+}
+
+export const resolveOgcTilesetResource = async (
+  initialUrl: string,
+  authHeaders: Record<string, string>,
+): Promise<ResolvedTileset> => {
+  // Prefer a known/remembered public S3 URL first — the OGC gateway currently
+  // hangs or returns 502 for long periods, which blocked the whole layer load.
+  const fallbackUrl = findFallbackUrl(initialUrl);
+  if (fallbackUrl !== null) {
```
Dead0utside, line 48: Is it intentional that the S3 URL is preferred over the OGC endpoint, rather than only being used as a fallback when OGC fails? The file-level comment and PR description describe this as a fallback, but for known/remembered URLs OGC is skipped entirely as long as S3 succeeds.

ui/src/features/layer/slice/tileset-slice-metadata.ts
Dead0utside, line 77: Is it guaranteed that sliceDirection/sliceNumber metadata is either present for all slice tiles or absent for all of them? As implemented, the presence of directional metadata on a single tile disables the URI fallback globally, so URI-only slices in a partially annotated tileset would be silently dropped.

ui/src/features/layer/slice/tileset-slice-metadata.ts
Dead0utside: Should we validate sliceNumber with Number.isFinite() / Number.isInteger() here? With the current typeof number === 'number' check, values like NaN and Infinity would still be accepted.

ui/src/features/layer/slice/tileset-slice-metadata.test.ts
Dead0utside, line 1: if the case with mixed types of metadata (see one of the previous comments) is realistic, it must be covered in tests

ui/src/features/layer/slice/ogc-tileset-resolve.ts
```diff
@@ -0,0 +1,213 @@
...
+  'swissgeol-ogc-api-cache-geometries.s3.eu-central-1.amazonaws.com';
+const RESOLVED_URL_STORAGE_PREFIX = 'swissgeol:ogc-tileset-resolved:';
+/** Give up on the OGC gateway quickly and use the public S3 cache instead. */
+const OGC_FETCH_TIMEOUT_MS = 8_000;
+
+/**
+ * Known public S3 tileset URLs used when the OGC API is unavailable.
+ * Keyed by OGC path suffix (`/collections/{id}/styles/{style}/download_format/tiles3d`).
+ */
+const KNOWN_S3_TILESET_FALLBACKS: Readonly<Record<string, string>> = {
```
Dead0utside, line 29: Should this dataset-specific fallback URL live in generic resolver code? It couples the slice infrastructure directly to collection 14279 / style 5 and a specific S3 object name. Could this come from layer/config metadata instead?

ui/src/features/layer/slice/ogc-tileset-resolve.ts
```diff
@@ -0,0 +1,213 @@
...
+    return null;
+  }
+};
+
+const rememberResolvedUrl = (initialUrl: string, resolvedUrl: string): void => {
+  if (typeof sessionStorage === 'undefined') {
+    return;
+  }
+  try {
+    if (!resolvedUrl.includes('amazonaws.com')) {
```
Dead0utside, line 130: Could we validate the hostname here rather than using includes('amazonaws.com')? Both this and rewriteOgcContentUrisToResolvedBase() would also accept unrelated hosts containing that string. Since S3_CACHE_HOST is already defined, an exact host check might be safer.

ui/src/features/layer/slice/ogc-tileset-resolve.ts
```diff
@@ -0,0 +1,213 @@
...
+    root?: TilesetNode;
+  };
+
+  const rewrite = (node: TilesetNode | undefined): void => {
+    if (node === undefined) {
+      return;
+    }
+    const uri = node.content?.uri;
+    if (typeof uri === 'string' && uri.includes(OGC_HOST)) {
+      const fileName = uri.split('/').pop();
```
Dead0utside, line 196: Is it guaranteed that all content objects are stored in the same S3 directory as the tileset JSON? The rewrite keeps only uri.split('/').pop(), so any nested content path would be flattened.

ui/src/features/layer/slice/ogc-tileset-resolve.ts
```diff
@@ -0,0 +1,213 @@
...
+    } catch (fallbackError) {
+      console.warn(
+        'Public S3 tileset fallback failed; trying OGC API:',
+        fallbackError,
+      );
+    }
+  }
+
+  try {
+    const resolved = await fetchTilesetJson(initialUrl, authHeaders);
```
Dead0utside, line 60: If the first S3 fallback attempt fails (currently line 47), do we want to retry the exact same URL again after OGC fails? For deterministic errors such as 403/404 this can add another 60s delay. With the current timeouts the worst-case failure path can take roughly 128s (S3 60s → OGC 8s → S3 60s).

ui/src/features/layer/slice/axis-tileset-slot.ts
```diff
@@ -0,0 +1,827 @@
...
+          next.originalJson,
+          next.baseUrl,
+          next.headers,
+          next.direction,
+          next.numbers,
+        );
+      }
+    } finally {
+      this.isUpdating = false;
+      for (const resolve of this.pendingResolvers.splice(0)) {
```
Dead0utside, line 189: I might be missing something in the concurrency flow here: if several slice changes happen quickly (A → B → C), pending is replaced with the latest selection, but all waiting `setSlices()` callers are resolved afterwards. Could the older `applySliceSelection`(A/B) calls then continue and call `prioritizeAxisNeighborhood()` / `warmAxisNeighborhood()` using their stale selections? If so, could this cause preloading to move back to an outdated neighbourhood after C has already become active?

ui/src/features/layer/slice/axis-tileset-slot.ts
```diff
@@ -0,0 +1,827 @@
...
+    generation: number,
+    originalJson: unknown,
+    baseUrl: string,
+    headers: Record<string, string>,
+    direction: OgcSliceDirection,
+    number: number,
+  ): Promise<boolean> {
+    const key = `${direction}:${number}`;
+    const existing = this.cache.get(key);
+    if (existing !== undefined && !existing.tileset.isDestroyed()) {
```
Dead0utside, line 248: Should the cached fast path in `warmOneSlice()` also check `tilesLoaded`? A tileset can already be cached after an interactive load timed out after 400 ms and then be frozen with `preloadWhenHidden = false`. In that case this returns true without actually finishing the warm-up, so revisiting the supposedly warmed slice may still require loading.

ui/src/features/layer/slice/axis-tileset-slot.ts
```diff
@@ -0,0 +1,827 @@
...
+   * slice it wants is already on screen.
+   */
+  private supersedeInFlight(): void {
+    this.loadGeneration += 1;
+    this.pending = null;
+  }
+
+  /** Hide this axis without destroying (used in Multiple mode for inactive axes). */
+  clear(): void {
+    this.loadGeneration += 1;
```
Dead0utside, line 342: Should clear() also increment warmGeneration? Currently clearing an axis cancels interactive loads via loadGeneration, but an already running warmSlices() can keep building/cache-warming slices for an axis that is no longer active.

ui/src/features/layer/slice/axis-tileset-slot.ts
Dead0utside, line 1: it appears that there are no tests for `axis-tileset-slot.ts`. I think the feature could benefit from some tests

ui/src/features/layer/slice/tileset-slice-prefetch.ts
```diff
@@ -0,0 +1,407 @@
...
+            isFetched = true;
+          }
+        } catch {
+          if (signal.aborted) {
+            this.inFlightUrls.delete(job.url);
+            return;
+          }
+        }
+        this.inFlightUrls.delete(job.url);
+        if (isFetched && !this.completedUrls.has(job.url)) {
```
Dead0utside, line 317: What should happen when a preload request fails? The URL is already included in `axisTotals`, but only successful requests increment `axisLoaded`, and failed jobs aren't retried/requeued here. Could one failed request leave the preload progress permanently below 100% while the HUD stays open?

ui/src/features/layer/slice/tileset-slice-prefetch.ts
```diff
@@ -0,0 +1,407 @@
...
+      while (!signal.aborted) {
+        const job = this.pending.shift();
+        if (job === undefined) {
+          return;
+        }
+        this.pendingUrls.delete(job.url);
+        this.inFlightUrls.add(job.url);
+        let isFetched = false;
+        try {
+          const response = await fetch(job.url, {
```
Dead0utside, line 298: Do we want a timeout for these background fetches? If one or more requests hang rather than fail, they can occupy preload workers indefinitely (potentially all 6) until the HUD is closed/aborted.

ui/src/features/layer/slice/tileset-slice-prefetch.ts
```diff
@@ -0,0 +1,407 @@
...
+          credentials: 'omit',
+          signal,
+          cache: 'force-cache',
+        });
+      } catch {
+        if (signal?.aborted) {
+          return;
+        }
+      }
+      done += 1;
```
Dead0utside, line 391: I noticed that failed requests are handled differently here than in `SlicePreloadQueue`: `prefetchUrls()` increments done even when the fetch fails, while `SlicePreloadQueue` only increments progress for successful requests. Is this difference intentional? It might be worth using the same progress semantics in both places.

ui/src/features/layer/slice/tileset-slice-prefetch.ts
Dead0utside, line 1: do you think we need tests here?

ui/src/features/layer/slice/tileset-slice-prune.ts
```diff
@@ -0,0 +1,243 @@
...
+    }
+    const childRegions = tile.children.map((child) => getRegion(child));
+    // Only tighten when every kept child exposes a region. Unioning a subset
+    // would produce a volume that culls the children we cannot account for
+    // (e.g. box/sphere bounding volumes).
+    if (
+      childRegions.length > 0 &&
+      childRegions.every((region): region is number[] => region !== null)
+    ) {
+      tile.boundingVolume = {
```
Dead0utside, line 110: Is it guaranteed that tiles with children never have their own content in these tilesets? If a kept tile has both content and children, this replaces its bounding volume with the union of the children only, which could exclude the tile's own content and potentially cause incorrect culling.

ui/src/features/layer/slice/tileset-slice-prune.test.ts
Dead0utside, line 1: if the previous comment is a possible case, please add a test case where the parent tile has both children and its own content

ui/src/features/layer/controllers/layer-tiles3d.controller.ts
```diff
@@ -1,79 +1,515 @@
...
+    }
+  }
+
+  private async addFullTileset(resource: Resource): Promise<void> {
+    this.destroyAxisSlots();
+    this.sliceMetadata = null;
+    this.defaultSliceSelection = null;
+
     const tileset = await Cesium3DTileset.fromUrl(resource, {
       show: true,
```
Dead0utside, line 584: Was it intentional to remove the existing tileset options here? This path is still used for normal/non-slice 3D-tiles layers, so settings like `enableCollision`, `maximumScreenSpaceError`, `cullRequestsWhileMoving`, `dynamicScreenSpaceError`, etc. would now fall back to Cesium defaults. The slice tilesets still seem to use these options in `AxisTilesetSlot`.

ui/src/features/layer/controllers/layer-tiles3d.controller.ts
```diff
@@ -109,19 +545,114 @@ export class Tiles3dLayerController extends BaseLayerController<Tiles3dLayer> {
...
+      const slot = this.axisSlots.get(axis);
+      if (slot === undefined) {
+        continue;
+      }
+      const numbers = numbersByAxis[axis] ?? [];
+      const numbersKey = JSON.stringify(numbers);
+      if (this.appliedNumbersByAxis.get(axis) === numbersKey) {
+        continue;
+      }
+      this.appliedNumbersByAxis.set(axis, numbersKey);
```
Dead0utside, line 644: Should `appliedNumbersByAxis` only be updated after `setSlices()` has successfully applied the selection? `AxisTilesetSlot` catches build failures internally and returns, so currently a failed load can still be recorded as applied here. A later attempt with the same selection would then be skipped by the equality check above.

ui/src/features/catalog/display/details/catalog-display-slice-detail.element.ts
```diff
@@ -0,0 +1,756 @@
...
+              </span>
+              ${this.renderStepper(value, min, max, (next) =>
+                this.setSingleSlice(axis, next),
+              )}
+            </div>
+            <ngm-core-slider
+              .value=${value}
+              .min=${min}
+              .max=${max}
+              .step=${1}
```
Dead0utside, line 408: Are slice numbers guaranteed to be contiguous? The slider uses min/max with step = 1, and the stepper also increments/decrements the numeric value by 1. If `getAxisNumbers()` can return something like [10, 20, 30], this would allow selecting slice numbers that don't actually exist. Would it be safer for the UI to step through the numbers array instead?

Dead0utside: this comment also potentially applies to 
`@click=${() => onChange(value - 1)}`
`@click=${() => onChange(value + 1)}`

ui/src/features/catalog/display/catalog-display-list-item.element.ts
```diff
@@ -69,20 +70,31 @@ export class CatalogDisplayListItem extends CoreElement {
...
+    // Tileset and slice metadata finish loading asynchronously after activation.
+    const startedAt = Date.now();
+    const checkInterval = setInterval(() => {
+      const isReady = !!controller.tileset?.boundingSphere;
+      if (isReady) {
+        this.canZoom = true;
+      }
+      if (
+        (isReady && Date.now() - startedAt > 2_000) ||
+        Date.now() - startedAt > 15_000
```
Dead0utside, line 92: What happens if the tileset becomes ready after this 15s timeout? At that point the polling stops while `canZoom` is still false, so it looks like the zoom button could remain disabled permanently. Could this be driven by readiness/state instead of stopping the check after a fixed timeout?

Dead0utside (changes requested):

ui/src/features/layer/slice/ogc-tileset-resolve.ts
uhon: Outdated — S3-fallback logic removed entirely; only a direct OGC fetch remains

ui/src/features/layer/slice/tileset-slice-metadata.ts
uhon: Fixed — mixed case now explicitly detected, warns, and omits unsafe tiles instead of silently defaulting

ui/src/features/layer/slice/tileset-slice-metadata.ts
uhon: Fixed — now uses Number.isInteger().

ui/src/features/layer/slice/tileset-slice-metadata.test.ts
uhon: Fixed — dedicated test added.

ui/src/features/layer/slice/ogc-tileset-resolve.ts
uhon: Outdated — that fallback code was removed.

ui/src/features/layer/slice/ogc-tileset-resolve.ts
uhon: Fixed — now does exact URL.host comparison.

ui/src/features/layer/slice/ogc-tileset-resolve.ts
uhon: Fixed — relative path now preserved via a path-prefix regex

ui/src/features/layer/slice/ogc-tileset-resolve.ts
uhon: Outdated — S3 retry path removed; single bounded (8s) OGC fetch only

ui/src/features/layer/slice/axis-tileset-slot.ts
uhon: Fixed — controller now checks isActive per axis before calling prioritize/warm.

ui/src/features/layer/slice/axis-tileset-slot.ts
uhon: Fixed — now checks existing.tileset.tilesLoaded.

ui/src/features/layer/slice/axis-tileset-slot.ts
uhon: Fixed — clear() now increments both loadGeneration and warmGeneration

ui/src/features/layer/slice/axis-tileset-slot.ts
uhon: Fixed — axis-tileset-slot.test.ts (271 lines) added

ui/src/features/layer/slice/tileset-slice-prefetch.ts
uhon: Fixed — bounded retries added, then counted done after max attempts.

ui/src/features/layer/slice/tileset-slice-prefetch.ts
uhon: Fixed — per-request timeout (SLICE_PREFETCH_REQUEST_TIMEOUT_MS) added.

ui/src/features/layer/slice/tileset-slice-prefetch.ts
uhon: Outdated — duplicate prefetchUrls implementation removed; only SlicePreloadQueue remains

ui/src/features/layer/slice/tileset-slice-prefetch.ts
uhon: Fixed — test file added

ui/src/features/layer/slice/tileset-slice-prune.ts
uhon: Fixed — explicit guard now skips recompute when a tile has its own content.

ui/src/features/layer/slice/tileset-slice-prune.test.ts
uhon: Fixed — corresponding test added.

ui/src/features/layer/controllers/layer-tiles3d.controller.ts
uhon: Fixed — options are present again.

ui/src/features/layer/controllers/layer-tiles3d.controller.ts
uhon: now only recorded once  setSlices()  resolves  true  (still active/succeeded), so a failed build no longer blocks retrying the same selection.

ui/src/features/catalog/display/details/catalog-display-slice-detail.element.ts
uhon: Non-contiguous slice numbers ( catalog-display-slice-detail.element.ts ): added  nearestSliceNumber / stepSliceNumber  helpers; slider drag and stepper +/- now snap to the closest actually-existing slice number instead of assuming a contiguous range.

ui/src/features/catalog/display/catalog-display-list-item.element.ts
uhon: polling now continues until the tileset is actually ready (or the controller changes/component disconnects) instead of giving up after a fixed 15s, so the zoom button no longer gets stuck disabled.

ui/src/features/layer/slice/axis-tileset-slot.ts
```diff
@@ -0,0 +1,872 @@
...
+   * Invalidate any in-flight or queued load: this selection is newer, and the
+   * slice it wants is already on screen.
+   */
+  private supersedeInFlight(): void {
+    this.loadGeneration += 1;
+    this.pending = null;
+  }
+
+  /** Hide this axis without destroying (used in Multiple mode for inactive axes). */
+  clear(): void {
```
Dead0utside, line 375: Should `clear()` also discard pending, similar to `supersedeInFlight()`? If a slice load is in progress and another selection is already queued, calling `clear()` invalidates the current load but leaves the queued one intact. Once the current `setSlices()` finishes, its queue-draining loop could still apply that older pending selection and make the axis visible again.

Dead0utside: Also, if this point is valid, write a regression test for it

ui/src/features/catalog/display/details/catalog-display-slice-detail.element.ts
```diff
@@ -0,0 +1,820 @@
...
+        this.bindPreloadProgress(this.controller);
+
+        if (this.controller?.supportsSliceSelection === true) {
+          this.activateHudPreload(this.controller);
+        }
+      }),
+    );
+  }
+
+  disconnectedCallback(): void {
```
Dead0utside, line 128: Closing the panel while a slider update is still pending might briefly re-trigger the layer subscription and turn HUD preloading back on after it was disabled. This depends on whether layerService.update() emits synchronously.

Would it make sense to do the sequence in this order?
```
this.cancelScheduledCommit();
this.commitDraftSingle();

this.hudActiveController?.setHudActive(false);
this.hudActiveController = null;

super.disconnectedCallback();
```

Dead0utside (approved):

PR #2048, opened 2026-08-17
Fix jquery integration (use by fomantic ui)

uhon:


Dead0utside (approved):
lgtm unless Copilot comments should be addressed

